PR #4, opened 2020-03-05
Polish the metrics handler

zcain117:
1. Change publish.sh to pass cluster and zone rather than hard-coding
2. Change metrics_handler main.py to use the new pubsub values
3. Rename cluster_id -> cluster_name in job_status_handler
4. Add postsubmit hook to re-deploy the metric handler when code changes
5. Add a README to the metric_handler dir to explain setup/config
6. Fix requirements.txt: most recent pip packages lead to deploy error

TESTED=Tested the commands in README.md and postsubmit.yaml for creating cloud function and cloud scheduler.

Change-Id: I33e3530dd2e819b0151ab45933568ba5b43fb2bf

PR #5, opened 2020-03-05
Polish metrics handler

zcain117:
1. Change publish.sh to pass cluster and zone rather than hard-coding
2. Change metrics_handler main.py to use the new pubsub values
3. Rename cluster_id -> cluster_name in job_status_handler
4. Add postsubmit hook to re-deploy the metric handler when code changes
5. Add a README to the metric_handler dir to explain setup/config
6. Fix requirements.txt: most recent pip packages lead to deploy error

TESTED=Tested the commands in README.md and postsubmit.yaml for creating cloud function and cloud scheduler.


will-cromar:
See my comments on #4

zcain117:
> See my comments on #4

Great catches in there, thanks. Updated

PR #15, opened 2020-03-11
Add test template for PyTorch/XLA C++ unit tests

zcain117:
Tested by pushing the 2 new generated yamls and running them.

Change-Id: Iec2b9537aa6e547d05c14195e14ba8d8c5833bd6

### DIFF
--- a/templates/pytorch/nightly/cpp-operations.libsonnet
+++ b/templates/pytorch/nightly/cpp-operations.libsonnet
@@ -23,7 +23,7 @@ local tpus = import "../../tpus.libsonnet";
       "bash",
       "pytorch/xla/test/cpp/run_tests.sh",
     ],
-    regressionTestConfig: {}
+    regressionTestConfig: null,
   },
   local v2_8 = {
     accelerator: tpus.v2_8,
--- a/templates/pytorch/nightly/cpp-operations.libsonnet
+++ b/templates/pytorch/nightly/cpp-operations.libsonnet
@@ -1,0 +1,39 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


templates/pytorch/nightly/cpp-operations.libsonnet
```diff
@@ -0,0 +1,39 @@
...
+  },
+  local v2_8 = {
+    accelerator: tpus.v2_8,
+  },
+  local v3_8 = {
+    accelerator: tpus.v3_8,
+  },
+
+  configs: [
+    operations + v2_8 + base.Functional + timeouts.Hours(4),
```
will-cromar: Does this actually take 4 hours to run?

templates/pytorch/nightly/cpp-operations.libsonnet
```diff
@@ -0,0 +1,39 @@
...
+local tpus = import "../../tpus.libsonnet";
+
+{
+  local operations = base.PyTorchTest {
+    modelName: "cpp-operations",
+    command: [
+      "bash",
+      "pytorch/xla/test/cpp/run_tests.sh",
+    ],
+    regressionTestConfig: {}
```
will-cromar: Would it be clearer to use `null` here instead of `{}`? It would make JSonnet throw an error if you were to accidentally try to override a field.

zcain117: Sure, sounds good. Switched to that and re-pushed the yamls and re-ran them to make sure the change works

templates/pytorch/nightly/cpp-operations.libsonnet
zcain117: It takes ~2hrs plus/minus some wiggle room. I wanted a timeout that would allow a meaningful retry if the first attempt fails for a flaky reason

PR #17, opened 2020-03-16
Tests for Pytorch Fairseq Roberta

zcain117:
TESTED=Pushed functional and convergence yamls to kubernetes and ran to completion


taylanbil:
Can we add "pretrain" to these test names? I'd like to distinguish this and fine-tuning.

zcain117:
> Can we add "pretrain" to these test names? I'd like to distinguish this and fine-tuning.

I have:

1. Renamed the template file and the test name
2. Renamed the generated files
3. Deleted the kubernetes workloads that had the old test name
4. Created new kubernetes workloads with the new requested test name
5. Deleted all rows in the metrics table that had the old test name

I think the name change is now done

PR #18, opened 2020-03-17
Change path to model garden ResNet.

will-cromar:


will-cromar:
Corresponding model garden change: https://github.com/tensorflow/models/commit/ad710aa18a7869913f8355ba5deb00e60b1836ca#diff-c74f86f208c466b66e5057a3c347cf78

PR #25, opened 2020-03-23
Create pytorch-1.5 test suite. Update pytorch Docker image create to …

zcain117:
…allow for new version

TESTED=Pushed all the configs to kubernetes and ran every test at least once

Change-Id: I82b3505cd384111f651a095fcf0b95703e9077e7


zcain117:
Shortened all the test names, regenerated the test configs, pushed all to kubernetes and ran and they LGTM. I also cleaned up BigQuery of the data from the previous test names

PR #32, opened 2020-03-26
Refactor metrics handler main.py and add tests

will-cromar:
Move a lot of business logic into `metrics.py` and add tests. Add simple tests for `main.py`. Use `absltest` in `util_test.py`.

Ran coverage and got the following results:

```
Name                    Stmts   Miss  Cover   Missing
-----------------------------------------------------
alert_handler.py           81     43    47%   55, 57-72, 75-79, 85-87, 90, 96, 98, 106, 122, 150, 154-191
job_status_handler.py      72     57    21%   45-89, 105-152
main.py                   234    155    34%   88-96, 100, 104, 133-143, 148-181, 199-254, 260-268, 296, 316-321, 324-328, 342, 354-420, 424-515
main_test.py               28      0   100%
metrics.py                114     17    85%   38-40, 56-57, 79, 81, 104, 107, 148, 158-160, 184, 200, 207, 222
metrics_test.py            60      0   100%
util.py                    26      1    96%   76
util_test.py               28      0   100%
-----------------------------------------------------
TOTAL                     643    273    58%
```

Coverage is still not great, but most of these numbers are up from ~0%. Tested manually by creating a new cloud function and playing back some PubSub messages.

### DIFF
--- a/metrics_handler/main.py
+++ b/metrics_handler/main.py
@@ -251,6 +251,7 @@ class CloudMetricsHandler(object):
             logs_link=self.stackdriver_logs_link)
 
   def get_metrics_history_from_bigquery(self):
+    """Returns the historic values of each metric for a given model."""
     query_result = self.bigquery_client.query(
         'SELECT * FROM `{}` WHERE test_name like \"{}\"'.format(
             self.metric_history_table_id, self.test_name)).result()
@@ -272,6 +273,7 @@ class CloudMetricsHandler(object):
 
     Args:
       job_status (dict): Contains information about the Kubernetes Job.
+      metrics_history(dict): Historic values of each metric.
       new_metrics(dict): Key is metric name and value is MetricPoint containing
         the latest aggregated value for that metric.
 
--- a/metrics_handler/main.py
+++ b/metrics_handler/main.py
@@ -115,7 +115,7 @@ class CloudMetricsHandler(object):
       raw_metrics = metrics.read_metrics_from_events_dir(
           self.events_dir, tags_to_ignore)
     except:
-      self.logger.warn(str(e))
+      self.logger.warning(str(e))
 
     default_aggregation_strategies = self.metric_collection_config.get(
         'default_aggregation_strategies')
--- a/metrics_handler/main.py
+++ b/metrics_handler/main.py
@@ -111,8 +111,11 @@ class CloudMetricsHandler(object):
     """
     tags_to_ignore = set(
         self.metric_collection_config.get('tags_to_ignore', []))
-    raw_metrics = metrics.read_metrics_from_events_dir(
-        self.events_dir, tags_to_ignore)
+    try:
+      raw_metrics = metrics.read_metrics_from_events_dir(
+          self.events_dir, tags_to_ignore)
+    except:
+      self.logger.warn(str(e))
 
     default_aggregation_strategies = self.metric_collection_config.get(
         'default_aggregation_strategies')
@@ -135,8 +138,11 @@ class CloudMetricsHandler(object):
                          'See README for how to set up the config.')
       tag = tta_config['accuracy_tag']
       threshold = tta_config['accuracy_threshold']
-      final_metrics['time_to_accuracy'] = metrics.time_to_accuracy(
-          raw_metrics, tag, threshold)
+      try: 
+        final_metrics['time_to_accuracy'] = metrics.time_to_accuracy(
+            raw_metrics, tag, threshold)
+      except ValueError as e:
+        self.logger.error(str(e))
 
     return final_metrics
 
--- a/metrics_handler/main.py
+++ b/metrics_handler/main.py
@@ -24,6 +24,7 @@ import uuid
 
 import alert_handler
 import job_status_handler
+import metrics
 
 import google.api_core.exceptions
 from google.cloud import bigquery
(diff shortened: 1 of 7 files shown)


metrics_handler/metrics.py
```diff
@@ -0,0 +1,246 @@
...
+          'comparison. The comparison was: {}'.format(comparison))
+    values = [v.metric_value for v in value_history]
+    mean = np.mean(values)
+    stddev = np.std(values)
+
+    if 'less' in comparison:
+      lower_bound = -math.inf
+      upper_bound = mean + (stddev * threshold.value)
+    elif 'greater' in comparison:
+      lower_bound = max(mean - (stddev * threshold.value), 0)
```
will-cromar: I copied this over from main.py. Is there a reason we can't allow values less than 0?

metrics_handler/main_test.py
```diff
@@ -0,0 +1,98 @@
...
+      events_dir=self.temp_dir,
+      stackdriver_logs_link=None,
+      metric_collection_config={
+        'default_aggregation_strategies': ['final', 'min',]
+      },
+      regression_test_config={},
+      test_type=None,
+      accelerator=None,
+      framework_version=None,
+      logger=self.logger, # pass in absl logging module
```
zcain117: For this comment (and the matching one below):
1. Formatting:  2 spaces between the comma and the #, Capitalize first letter, end with period
2. Prefix with TODO if this is something we want to do. Right now we're passing AlertHandler, but the comment refers to absl logging, so I'm assuming this is a future intended change

metrics_handler/metrics.py
```diff
@@ -0,0 +1,260 @@
...
+
+  Args:
+    raw_metrics: dict mapping TensorBoard tags to list of MetricPoint.
+    default_strategies: list of aggregation strategies to use as the default.
+    metric_strategies: dict mapping tags to aggregation strategies.
+
+  Returns:
+    dict mapping metric name to MetricPoint.
+  """
+  if not raw_metrics:
```
zcain117: I think it's better to return empty dict and log a warning for empty raw_metrics since not all models will write to Tensorboard.

If we have raw_metrics but no `default_strategies`, I could go either way on warning vs. error but probably would prefer logging an error and continuing the run. That second change can be later though

metrics_handler/metrics.py
```diff
@@ -0,0 +1,260 @@
...
+  Available aggregation strategies: `final`, `min`, `max`.
+
+  Args:
+    raw_metrics: dict mapping TensorBoard tags to list of MetricPoint.
+    default_strategies: list of aggregation strategies to use as the default.
+    metric_strategies: dict mapping tags to aggregation strategies.
+
+  Returns:
+    dict mapping metric name to MetricPoint.
+  """
```
zcain117: Add the `Raises` section to the docstring

metrics_handler/main.py
```diff
@@ -206,59 +112,38 @@ def get_metrics_from_events_dir(self):
...
-            metrics, strategy)
-
+    raw_metrics = metrics.read_metrics_from_events_dir(
+        self.events_dir, tags_to_ignore)
+
+    default_aggregation_strategies = self.metric_collection_config.get(
+        'default_aggregation_strategies')
+    metric_to_aggregation_strategies = self.metric_collection_config.get(
+        'metric_to_aggregation_strategies')
+    final_metrics = metrics.aggregate_metrics(
```
zcain117: Right now I don't think this would translate to any alerts if it fails since it will raise a ValueError and the process will terminate. We can either pass a logger to all of the `metrics.*` methods and have those methods decide which failure are email-worthy or wrap all calls with try/catch and log the errors in this method like you do below with metrics.time_to_accuracy

metrics_handler/main_test.py
will-cromar: Oops, that's an old comment. Removed.

metrics_handler/metrics.py
will-cromar: Good idea. Done.

metrics_handler/metrics.py
zcain117: It just affects how the bounds are rendered in the dashboard and I couldn't think of any metrics where it would be relevant to have negative numbers as bounds. That said, now I think it's better to remove this `max` and push any necessary changes for rendering onto the dashboard, e.g. the dashboard could replace negative values with 0 if that's what we want

metrics_handler/metrics.py
will-cromar: Done.

metrics_handler/main.py
will-cromar: It will fall through and get caught/logged here: 

```
    try:
      logger.info('Pubsub message to process: {}'.format(msg))
      should_ack = _process_pubsub_message(msg, status_handler, logger)
    except Exception as e:
      logger.error(
          'Encountered exception `{}` while attempting to '
          'process message. The message will be acknowledged to prevent more '
          'crashes. The message was: {}'.format(e, msg))
      should_ack = True
```

I did add a try/catch to add more context to the error before re-raising it.

metrics_handler/metrics.py
will-cromar: I agree that we can push that logic to the dashboard. Done.

PR #33, opened 2020-03-27
TF2.x Keras API tests on TPU

gagika:
Adding TF 2.x Keras API tests on TPUs.

### DIFF
--- a/templates/garden/nightly/keras-api.libsonnet
+++ b/templates/garden/nightly/keras-api.libsonnet
@@ -1,0 +1,111 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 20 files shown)


templates/garden/nightly/keras-api.libsonnet
```diff
@@ -0,0 +1,111 @@
...
+{
+  local base_command = ||| 
+    export PATH=$PATH:/root/google-cloud-sdk/bin &&
+    gcloud source repos clone cloudtpu-tf20-api-tests --project=gcp-tpupods-demo &&
+    cd cloudtpu-tf20-api-tests &&
+    pip3 install behave &&
+  |||,
+
+  local keras_test = base.GardenTest {
+    testFeature:: "aaa",
```
zcain117: just to confirm, 'aaa' is what you want here?

gagika: Keras API first test feature name is "aaa_connection"  adding "aaa" will run that (by matching prefix), so by default I am setting that test feature, but all test cases right now override this value.  
behave framework run different test features alphabetical order, we named it with "aaa" to make sure it's first test, and that test basically check connectivity with TPU device.

PR #34, opened 2020-03-27
Create new mixin mode and switch keras-api tests to use it.

zcain117:
This will help offset the TPU usage spike that we get by running many functional tests all at the same time.

Change-Id: Ic4faf0084c19d879a7a9ed1a7bef292d31e94f95

### DIFF
--- a/templates/mixins.libsonnet
+++ b/templates/mixins.libsonnet
@@ -24,15 +24,6 @@ local timeouts = import "timeouts.libsonnet";
       preemptible: true,
     },
   },
-  Micro:: {
-    mode: "micro",
-    timeout: timeouts.one_hour,
-    # Run at 2AM PST daily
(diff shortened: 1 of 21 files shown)


templates/mixins.libsonnet
```diff
@@ -24,6 +24,15 @@ local timeouts = import "timeouts.libsonnet";
       preemptible: true,
     },
   },
+  Micro:: {
+    mode: "micro",
+    timeout: timeouts.one_hour,
```
will-cromar: I think we can set a more aggressive timeout, e.g. 10 minutes

will-cromar: Nevermind, these tests are longer than I thought.

gagika: some tests run longer than 10 minutes, and some test causes run many different combinations of input parameters

PR #40, opened 2020-03-30
Skip regression check if `regression_test_config` is null

will-cromar:


metrics_handler/main.py
```diff
@@ -415,9 +413,11 @@ def _process_pubsub_message(msg, status_handler, logger):
       regression_test_config, test_type, accelerator, framework_version, logger)
 
   new_metrics = handler.get_metrics_from_events_dir()
-  metrics_history = handler.get_metrics_history_from_bigquery()
-  metric_name_to_visual_bounds = handler.compute_bounds_and_report_errors(
-      job_status, metrics_history, new_metrics)
+  if self.regression_test_config:
+    metrics_history = handler.get_metrics_history_from_bigquery()
+    metric_name_to_visual_bounds = handler.compute_bounds_and_report_errors(
```
zcain117: I think initialize this None before this `if` block, otherwise you'll get a name error below when passing this name to `add_status_and_metrics_to_bigquery`. Should also verify that `add_status_and_metrics_to_bigquery` does the right thing if given None for the 3rd arg

will-cromar: Good catch. Done.

PR #41, opened 2020-03-31
Add unit tests for job_status_handler. Compute wall time using publis…

zcain117:
…h time.

Change-Id: I4807df89c7d68d620d39b1118bb9753fe16879b0

### DIFF
--- a/metrics_handler/job_status_handler_test.py
+++ b/metrics_handler/job_status_handler_test.py
@@ -1,0 +1,116 @@
+# Lint as: python3
+"""Tests for job_status_handler."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import datetime
+import time
+import types
+
+from absl.testing import absltest
+import alert_handler
+import job_status_handler
+
+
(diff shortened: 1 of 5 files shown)


metrics_handler/job_status_handler_test.py
```diff
@@ -0,0 +1,116 @@
...
+  def setUp(self):
+    self.logger = alert_handler.AlertHandler(
+      project_id=None,
+      write_to_logging=True,
+      write_to_error_reporting=False,
+      write_to_email=False)
+    self.handler = job_status_handler.JobStatusHandler(
+        "unused", "unused", "unused", self.logger)
+
+  def test_interpret_status_success_with_completion_time(self):
```
will-cromar: Could these tests be done as parameterized tests? e.g. https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/metrics_handler/metrics_test.py#L142

zcain117: Switched to parameterized, PTAL @will-cromar

PR #42, opened 2020-03-31
Add config option to disable email alerts for crashes/timeouts on a per-test basis.

zcain117:
Update README to document the new option. Use the new option to disable emails for keras-api tests.

### DIFF
--- a/templates/garden/nightly/keras-api.libsonnet
+++ b/templates/garden/nightly/keras-api.libsonnet
@@ -33,6 +33,9 @@ local tpus = import "../../tpus.libsonnet";
       "-c",
       base_command + " behave -e ipynb_checkpoints --tags=-fails -i %s" % self.testFeature,
     ],
+    regressionTestConfig: {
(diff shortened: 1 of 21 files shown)


templates/garden/nightly/keras-api.libsonnet
```diff
@@ -33,6 +33,9 @@ local tpus = import "../../tpus.libsonnet";
       "-c",
       base_command + " behave -e ipynb_checkpoints --tags=-fails -i %s" % self.testFeature,
     ],
+    regressionTestConfig: {
+      alert_for_failed_jobs: false,
```
zcain117: @will-cromar `false` or `False`? We use lowercase [here](https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/templates/base.libsonnet#L41) but that could also be wrong

will-cromar: Lower-case is correct https://jsonnet.org/ref/spec.html

PR #43, opened 2020-03-31
Generate readable multi-line YAML strings for metrics config

will-cromar:
Also generate YAML for oneshot tests.


templates/base.libsonnet
```diff
@@ -81,7 +81,7 @@ local tpus = import "tpus.libsonnet";
                   test_name: config.testName,
                   metric_collection_config: config.metricCollectionConfig,
                   regression_test_config: config.regressionTestConfig,
-                }, " "),
+                }, " ") + "\n",  // Add newline to make JSonnet generate a multi-line YAML string
```
zcain117, line 84: End comment with period

will-cromar: Done.

PR #45, opened 2020-03-31
Add utility function to better support multi-command tests.

will-cromar:
Update PyTorch RoBERTa and TF Keras API tests.

No update to RoBERTa generated files because I accidentally committed them in #43.

Generated YAML files created with this new function will have nicely-formatted multi-line strings instead of having an escaped string packed onto one line.

### DIFF
--- a/templates/pytorch/nightly/roberta-pre.libsonnet
+++ b/templates/pytorch/nightly/roberta-pre.libsonnet
@@ -21,7 +21,7 @@ local utils = import "../../utils.libsonnet";
   local roberta = base.PyTorchTest {
     modelName: "roberta-pre",
     paramsOverride: {
-      maxEpochs: error "Must set `maxEpoch`",
+      maxEpoch: error "Must set `maxEpoch`",
     },
     command: utils.scriptCommand(
       |||
--- a/templates/pytorch/nightly/roberta-pre.libsonnet
+++ b/templates/pytorch/nightly/roberta-pre.libsonnet
@@ -15,16 +15,48 @@
 local base = import "base.libsonnet";
 local timeouts = import "../../timeouts.libsonnet";
 local tpus = import "../../tpus.libsonnet";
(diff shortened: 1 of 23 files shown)


templates/pytorch/nightly/roberta-pre.libsonnet
```diff
@@ -15,16 +15,48 @@
...
-  local base_command = "pip install --editable tpu-examples/deps/fairseq && python3 /tpu-examples/deps/fairseq/train.py /datasets/wikitext-103 --tensorboard-logdir=$(MODEL_DIR) --task=masked_lm --criterion=masked_lm --arch=roberta_base --sample-break-mode=complete --tokens-per-sample=512 --optimizer=adam --adam-betas='(0.9,0.98)' --adam-eps=1e-6 --clip-norm=0.0 --lr-scheduler=polynomial_decay --lr=0.0005 --warmup-updates=10000 --dropout=0.1 --attention-dropout=0.1 --weight-decay=0.01 --update-freq=16 --log-format=simple --train-subset=train --valid-subset=valid --num_cores=8 --metrics_debug --save-dir=checkpoints --log_steps=30 --skip-invalid-size-inputs-valid-test --suppress_loss_report --input_shapes 16x512 18x480 21x384",
-
   local roberta = base.PyTorchTest {
     modelName: "roberta-pre",
-    command: [
-      "/bin/bash",
-      "-c",
-    ],
+    paramsOverride: {
+      maxEpochs: error "Must set `maxEpoch`",
```
zcain117, line 24: The name before the `:` says `maxEpochs` but the error message says `maxEpoch` and the variable in the format string also says `maxEpoch` and overrides below use `maxEpoch`. I don't know why this is able to work as-is but I'd switch the name to `maxEpoch`

will-cromar: Nice catch, fixed.

PR #50, opened 2020-04-01
Lowercase the testnames for Resnet-MP.

taylanbil:


taylanbil:
Renamed file, did not gen the yamls again as there should be no need.

PR #53, opened 2020-04-02
Fix pytorch test targets to use lowercase filename.

zcain117:


zcain117:
The roberta yaml diffs are unrelated. I forgot to submit them in https://github.com/GoogleCloudPlatform/ml-testing-accelerators/pull/49

PR #56, opened 2020-04-03
Add experimental PyTorch MNIST TPU pod test

will-cromar:


will-cromar:
To test this out, run `jsonnet templates/oneshot.jsonnet -S --tla-str test=pt-nightly-mnist-pods-experimental-pods | kubectl create -f -`

jysohn23:
Nice! Thanks 😄

will-cromar:
Updates since I published this draft:

- Use TPU provided by Kubernetes, which lets us queue jobs when no quota is available.
- Create base template for pods tests that hides the complexity of managing instance groups from the user and exposes the same interface as the other template.
- Use the smaller `google/cloud-sdk:slim` image as a base image for PyTorch pods tests, since we only use `gcloud` on the main K8s Pod anyway.
- Explicitly wait for instance group to initialize in setup script.
- Lower timeout for setup script since the TPU is already READY when the K8s pod starts.

will-cromar:
Tested: [oneshot link](https://pantheon.corp.google.com/kubernetes/job/europe-west4-a/europe-oneshots/default/pt-nightly-mnist-convergence-v2-32-cmbhv?project=xl-ml-test&tab=details&duration=P30D&pod_summary_list_tablesize=20&service_list_datatablesize=20)

images/pytorch-pods/entrypoint.sh
```diff
@@ -13,24 +13,36 @@
...
+export master=$(gcloud compute instance-groups \
+  list-instances \
+  ${INSTANCE_GROUP_NAME} \
+  --zone=${ZONE} \
+  --format="value(NAME)" \
+  --limit=1) && \
+gcloud -q compute ssh --internal-ip --zone=$ZONE $master \
+  --command "source /anaconda3/etc/profile.d/conda.sh && \
+  conda activate torch-xla-nightly && \
+  python -m torch_xla.distributed.xla_dist --tpu=$TPU_POD_NAME --conda-env=torch-xla-nightly -- $*"
```
jysohn23, line 44: This way for PT/TPU Pods tests jsonnet configs, we're only able to specify the `python $SCRIPT $ARGS` without the `xla_dist` module flags? I think we need the ability to specify other flags like `--env` (it's an `xla_dist` module flag) and also switch out the `--conda-env` versions to like `torch-xla-1.5` or smth.

Can we build the `python -m torch_xla...` full command from jsonnet?

taylanbil:
Awesome!

images/pytorch-pods/entrypoint.sh
```diff
@@ -13,24 +13,36 @@
...
+mkdir -p /root/google-cloud-sdk/bin/
+ln -s $(which gcloud) /root/google-cloud-sdk/bin/gcloud
 source /publish.sh
 
-# "$@" should look like python -m torch_xla.distributed.xla_dist --tpu=...
-docker-entrypoint.sh "$@"
-
-# Teardown resources
-/teardown-pytorch-pods.sh
+timeout 180 /setup-instances.sh && \
```
zcain117, line 34: If `setup-instances.sh` fails, will the overall Job fail? And would the logs from `entrypoint.sh`, `setup-instances.sh`, and `teardown-instances.sh` appear in stackdriver?

will-cromar: Yes and yes. I save the exit code from this chain of commands, unconditionally tear down any resources, and then return that exit code.

images/pytorch-pods/entrypoint.sh
will-cromar: Added template parameters for conda env and `xla_dist` flags. [oneshot link](https://pantheon.corp.google.com/kubernetes/job/europe-west4-a/europe-oneshots/default/pt-nightly-mnist-convergence-v2-32-nfnsj?project=xl-ml-test&tab=details&duration=P30D&pod_summary_list_tablesize=20&service_list_datatablesize=20)

PR #60, opened 2020-04-03
Set msg_publish_time column. Use None temporarily.

zcain117:
Sometimes pubsub delays can lead to duplicate rows in Bigquery. msg_publish_time will be used to avoid writing duplicate rows and to clean up duplicate rows


zcain117:
I have already updated the schema of our table to match this PR

PR #62, opened 2020-04-06
Add mixin for experimental tests.

will-cromar:
This can be used to disable autometed runs for test configs.

Disable some consistently failing tests.


templates/pytorch/nightly/python-ops.libsonnet
```diff
@@ -33,7 +34,7 @@ local tpus = import "../../tpus.libsonnet";
   },
 
   configs: [
-    operations + v2_8 + base.Functional + timeouts.Hours(2),
-    operations + v3_8 + base.Functional + timeouts.Hours(2),
+    operations + v2_8 + base.Functional + timeouts.Hours(2) + mixins.Experimental,
```
zcain117, line 37: Let's not use experimental for these tests. It's important to get the new run daily for these. Jack and Davide have been making progress and this test is our first line of defense to know something has broken. They've fixed it a few times and it just happens to fail with new errors after each fix

will-cromar: Understood. Fixed.

PR #67, opened 2020-04-06
Change deployment manager config to template with optional pools.

will-cromar:


will-cromar:
Default cluster config should be much cheaper to run now. Also add a multi-GPU pool to config.

deployments/README.md
```diff
@@ -2,10 +2,13 @@
 
 Deployment Manager docs: https://cloud.google.com/deployment-manager/docs/
 
-To create your testing cluster, run the following:
+To create testing cluster, run one of the followiing commands:
```
zcain117, line 5: typo

will-cromar: fixed

PR #70, opened 2020-04-10
Publish PubSub message from an initContainer

will-cromar:
Now you can use any image that has a model that writes to GCS without having to install `gcloud` or add `publish.sh` 🎊

### DIFF
--- a/images/publisher/publish.sh
+++ b/images/publisher/publish.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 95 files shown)


templates/base.libsonnet
```diff
@@ -63,8 +63,67 @@ local tpus = import "tpus.libsonnet";
...
+              valueFrom: {
+                fieldRef: {
+                  fieldPath: "metadata.labels['job-name']",
+                },
+              },
+            },
+            {
+              name: "MODEL_DIR",
+              value: 
+                "gs://xl-ml-test-us-central1/k8s/%(modelName)s/%(mode)s/%(acceleratorName)s/$(JOB_NAME)" % config,
```
zcain117, line 102: `us-central1` is hardcoded here but we already have a oneshots cluster in europe. It seems to me this wouldn't hard-break anything really: I guess our europe clusters might have longer read/write time to GCS. Would it cause any problems you can think of?

will-cromar: For now, it will write summaries and checkpoints more slowly, which shouldn't be a huge deal. All of our automated tests still run in us-central1 for the moment. Long term, we should move this bucket name to a cluster-level ConfigMap to enable tests in other regions.

images/publisher/publish.sh
```diff
@@ -23,4 +27,7 @@ runtime_vars="{
   \"cluster_name\": \"$CLUSTER\"
 }"
 pubsub_message=`echo ${runtime_vars} | jq ". + ${METRIC_CONFIG}"`
-/root/google-cloud-sdk/bin/gcloud pubsub topics publish metrics-written --message "$pubsub_message"
+echo "Pubsub message: ${pubsub_message}"
+
+set -x
+gcloud pubsub topics publish metrics-written --message "$pubsub_message"
```
zcain117, line 33: doesn't need to be this PR but wanted to start you thinking about it: we might want a parallel pubsub for oneshot messages with a oneshot metrics handler listening on it. Then people could use oneshots ensure that their regression_test_config is WAI. So this topic name would need to be overrideable at the oneshot creation layer

will-cromar:
Tested e2e. Merging.

PR #72, opened 2020-04-11
Add tests for Huggingface GLUE tasks

jysohn23:
Models:
* `bert-base-cased`
* `distilbert-base-uncased`
* `roberta-large`
* `xlm-mlm-en-2048`
* `xlnet-large-cased`

Either this PR or https://github.com/GoogleCloudPlatform/ml-testing-accelerators/pull/71 will need to be rebased and resolve conflicts.

### DIFF
--- a/templates/pytorch/nightly/hf-glue.libsonnet
+++ b/templates/pytorch/nightly/hf-glue.libsonnet
@@ -1,0 +1,225 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+local base = import "base.libsonnet";
+local timeouts = import "../../timeouts.libsonnet";
+local tpus = import "../../tpus.libsonnet";
+local utils = import "../../utils.libsonnet";
+
+{
+  local command_common = |||
+    git clone https://github.com/huggingface/transformers.git
+    cd transformers && pip install .
+    git log -1
+    python examples/run_tpu_glue.py \
+      --tensorboard_logdir=$(MODEL_DIR) \
+      --task_name MNLI \
+      --do_train \
+      --do_eval \
+      --data_dir /datasets/glue/MNLI \
+      --max_seq_length 128 \
+      --learning_rate 3e-5 \
+      --num_train_epochs 3.0 \
+      --output_dir MNLI \
+      --overwrite_output_dir \
+      --logging_steps 50 \
+      --save_steps 1000 \
+      --num_cores=8 \
+      --only_log_master \
+      --metrics_debug \
+  |||,
+  local bert_base_cased = base.Convergence {
+    modelName: "hf-glue-bert-b-c",
+    command: utils.scriptCommand(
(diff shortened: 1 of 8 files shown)


templates/pytorch/nightly/hf-glue.libsonnet
```diff
@@ -0,0 +1,225 @@
...
+# limitations under the License.
+
+local base = import "base.libsonnet";
+local timeouts = import "../../timeouts.libsonnet";
+local tpus = import "../../tpus.libsonnet";
+local utils = import "../../utils.libsonnet";
+
+{
+  local command_common = |||
+    git clone https://github.com/huggingface/transformers.git
```
will-cromar, line 22: Would it be better to create a new image that has this repository or add it to our PyTorch image?

templates/pytorch/nightly/hf-glue.libsonnet
```diff
@@ -0,0 +1,225 @@
...
+    cd transformers && pip install .
+    git log -1
```
will-cromar, line 24: What does this line do?

templates/pytorch/nightly/hf-glue.libsonnet
jysohn23: I did initially do this here https://github.com/pytorch/xla/pull/1892 but based on comments, will clone as part of the test.

templates/pytorch/nightly/hf-glue.libsonnet
jysohn23: Just wanted to log the exact huggingface/transformers:master commit at which we're running for debugging purposes.

templates/pytorch/nightly/hf-glue.libsonnet
```diff
@@ -0,0 +1,225 @@
...
+      |||
+        %(common)s --model_type bert \
+        --model_name_or_path bert-base-cased \
+        --train_batch_size 32 \
+        --eval_batch_size 32
+      ||| % {common: command_common}
+    ),
+    regressionTestConfig+: {
+      metric_success_conditions+: {
+        "mnli/acc": {
```
zcain117, line 54: All of these metric names should be suffixed with "_final" if you are using default aggregation strategy. You could also use "max" but then you'd need to change your default aggregation strategy.

From [the README](https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/metrics_handler/README.md#regression_test_config):

```
  # (Required) Defines the success condition for each training metric. If any
  # metric does not meet the success condition, it will trigger an alert.
  #
  # Key for each entry is the metric name. Note that most metric names will
  # take the form of "<tensorboard tag>_<aggregration strategy>". Use the
  # special key "default" to set a success condition that will be used for any
  # metric that does not have an explicit success condition.
  #
  # NOTE: All metrics captured by metric_success_conditions will compute
  # lower/upper bounds and write them to BigQuery so that these bounds
  # can be used as a visual aid when viewing metric history. You have the
  # option of disabling alerts for any or all of these metrics with the other
  # optional fields described below.
  "metric_success_conditions": {
    "Accuracy/test_final": {
      "comparison": "greater",
      "success_threshold": {
        # Compare the metric value to a fixed value. For this type of
        # threshold, comparison can be "greater", "less", or "equal".
        "fixed_value": 99.0
      }
    },
    "default": {
      "comparison": "less",
      "success_threshold": {
        # Compare the metric to historical performance. For this type
        # of threshold, comparison can be "greater" or "less".
        # "greater": assert metric_value > (mean - 4.0 * stddev)
        # "less": assert metric_value < (mean + 4.0 * stddev)
        "stddevs_from_mean": 4.0
      },
      # (Optional) Wait for N points of data before enforcing this success
      # condition so that mean/stddev are more reliable. If not set, alerts
      # will probably fire on the first run since the standard deviation
      # will be 0.0.
      "wait_for_n_points_of_history": 10
    }
  }
```

PR #75, opened 2020-04-15
Rename 'functional' and 'convergence' in metrics handler.

will-cromar:


will-cromar:
Agreed. I will do that next week.

PR #79, opened 2020-04-15
Dashboard initial

zcain117:
Initial check in of the Bokeh-based dashboard for viewing pass/fail of tests and metrics history of tests.

TESTED=Ran with Google App Engine Flex as well as locally.


dashboard/modules/metric_history.py
```diff
@@ -0,0 +1,207 @@
...
+INNER JOIN `xl-ml-test.metrics_handler_dataset.job_history` AS job
+ON
+  metrics.uuid = job.uuid
+ORDER BY
+  run_date DESC
+"""
+
+
+def fetch_data(test_name):
+  dataframe = utils.run_query(QUERY % {'test_name': test_name},
```
will-cromar, line 75: Can you use BQ parameterized queries here (and for other queries) to avoid SQL injection concerns?

https://googleapis.dev/python/bigquery/latest/usage/queries.html#run-a-query-using-a-named-query-parameter
https://googleapis.dev/python/bigquery/latest/usage/queries.html#run-a-query-using-a-named-query-parameter

dashboard/utils.py
```diff
@@ -0,0 +1,66 @@
...
+      else:
+        df = _run(query)
+        redis_client.set(cache_key, df.to_json(orient='records'), ex=expire)
+    return df
+
+
+# TODO: Reconcile this with the similar helper method under
+# metrics_handler/util.py
+LOGS_DOWNLOAD_COMMAND = """gcloud logging read 'resource.type=k8s_container resource.labels.project_id={project} resource.labels.location={zone} resource.labels.cluster_name={cluster} resource.labels.namespace_name={namespace} resource.labels.pod_name:{pod}' --limit 10000000000000 --order asc --format 'value(textPayload)' --project={project} > {pod}_logs.txt && sed -i '/^$/d' {pod}_logs.txt"""
+LOG_LINK_REGEX = re.compile('https://console\.cloud\.google\.com/logs\?project=(\S+)\&advancedFilter=resource\.type\%3Dk8s_container\%0Aresource\.labels\.project_id\%3D(?P<project>\S+)\%0Aresource\.labels\.location=(?P<zone>\S+)\%0Aresource\.labels\.cluster_name=(?P<cluster>\S+)\%0Aresource\.labels\.namespace_name=(?P<namespace>\S+)\%0Aresource\.labels\.pod_name:(?P<pod>\S+)(\&dateRangeUnbound=backwardInTime)?')
```
will-cromar, line 56: I really don't like parsing the log link like this. If this is unavoidable, can you file a bug to include whatever extra information you need in the BigQuery table or original PubSub message?

dashboard/javascript_utils.py
```diff
@@ -0,0 +1,152 @@
+# Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
zcain117: I'll add a TODO here to investigate that

dashboard/modules/metric_history.py
zcain117: Switched to `pandas-gbq` parameterized queries here and in `main_heatmap.py`

zcain117:
Tested the latest changes running a local bokeh server

PR #83, opened 2020-04-20
Store workload link and logs download command in bigquery.

zcain117:
Once the PR is approved, I will update the Bigquery schema from GCP UI and submit this PR just after. I'll do this ~2-3 minutes after a run of the metrics handler, which runs every 15 min.


zcain117:
Just added the new columns

PR #88, opened 2020-04-21
Add region option to accelerator specs.

will-cromar:
Change output directory for generated files.

This PR is a no-op as far as our Kubernetes cluster is concerned, and it will unblock future work to enable support for splitting tests across clusters.


templates/README.md
```diff
@@ -15,14 +15,10 @@ jsonnet templates/oneshot.jsonnet -S --tla-str test=$TEST_NAME | kubectl create
...
-jsonnet templates/multifile.jsonnet -S -m k8s/gen
+jsonnet templates/multifile.jsonnet -S -m k8s/
 ```
 
-This command will output Kubernetes `CronJob` resources into [`k8s/gen`](../k8s/gen) directory. You can deploy these resources by running the following command:
-
-```bash
-kubectl apply -f k8s/gen -f k8s/
-```
+This command will output Kubernetes `CronJob` resources into [`k8s/gen`](../k8s/gen) directory.
```
zcain117, line 22: The dir will now be something like `k8s/zone/gen` so I think this link will become 404. Maybe direct the user just to `../k8s`

will-cromar: Done.

PR #95, opened 2020-04-22
Overwrite cache when running HF tests

jysohn23:


jysohn23:
Actually not quite, let me correct myself. This should just be a safeguard but not really change much in terms of behavior.

zcain117:
> Actually not quite, let me correct myself. This should just be a safeguard but not really change much in terms of behavior.

I see - so maybe it's just good practice to overwrite the cache to avoid spurious test failures? But in this case it won't fix the underlying problem of why the test is failing?

jysohn23:
Yes, this is orthogonal to the root cause fix.

zcain117:
Adding this flag in #101 since the generated files in this PR are out of date now

PR #99, opened 2020-04-23
Update PyTorch models to use PersistentVolumeClaim

will-cromar:
The PersistentVolumeClaim resource can be configured at the cluster level to point to a resource that is available in the zone of that cluster. I added configuration for a Cloud Filestore instance.

### DIFF
--- a/k8s/us-central1/gen/pt-nightly-cifar-tv-conv-v3-8.yaml
+++ b/k8s/us-central1/gen/pt-nightly-cifar-tv-conv-v3-8.yaml
@@ -70,6 +70,7 @@
             "volumeMounts":
             - "mountPath": "/dev/shm"
               "name": "dshm"
+              "readOnly": false
             - "mountPath": "/datasets"
               "name": "cifar-pd"
               "readOnly": true
--- a/k8s/us-central1/volumes.yaml
+++ b/k8s/us-central1/volumes.yaml
@@ -1,0 +1,25 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: pytorch-datasets
+spec:
+  capacity:
+    storage: 1T
+  accessModes:
+  - ReadOnlyMany
+  nfs:
+    path: /pytorch_datasets
+    server: 10.166.46.250
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: pytorch-datasets-claim
+spec:
+  accessModes:
+  - ReadOnlyMany
+  storageClassName: ""
+  volumeName: pytorch-datasets
+  resources:
+    requests:
+      storage: 1Ki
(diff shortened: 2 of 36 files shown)


templates/volumes.libsonnet
```diff
@@ -0,0 +1,73 @@
...
+  },
+
+  // Combines a map of VolumeSpec into a single pod spec mixin.
+  combinedMixin(volumeMap): std.foldl(
+    function(next, rest) rest + next,
+    [volumeMap[v].PodSpecMixin for v in std.objectFields(volumeMap)],
+    { }
+  ),
+
+  MemoryVolumeSpec:: self.VolumeSpec {
```
zcain117, line 52: Is this being used anywhere?

I see `dshm` changes above, but only to make `readonly` False.

It's fine to keep it even if unused, I'm just curious

k8s/us-central1/gen/pt-nightly-cifar-tv-conv-v3-8.yaml
```diff
@@ -70,6 +70,7 @@
             "volumeMounts":
             - "mountPath": "/dev/shm"
               "name": "dshm"
+              "readOnly": false
```
zcain117, line 73: could you file bugs for any models that had errors and still need to be migrated to the new dataset strategy?

k8s/us-central1/volumes.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+kind: PersistentVolume
```
zcain117, line 2: where in this PR do we point the volume to the PD? (Also: we still use PD?)

Could you file a bug to write a guide similar to [this](https://docs.google.com/document/d/19NEyXYR7HN12gCJ0uOg6TBe4GVW_3rq6xbBjFIJx7_A/edit#heading=h.jz0jr8sninst) on how a user would create a dataset storage vehicle to interact with this new volume stuff

templates/volumes.libsonnet
will-cromar: This is being used in `templates/pytorch/base.libsonnet`. `readOnly: false` was the default being used before. I figured it was more trouble than it's worth to avoid printing that default explicitly in this case.

k8s/us-central1/volumes.yaml
will-cromar: The `nfs` field below has the IP and fileshare name that we use for this cluster. I will add a note to b/152724749 to include details on how to add datasets to the Filestore instance. None of the tests using `PersistentVolumeClaim`s are using PDs anymore.

k8s/us-central1/gen/pt-nightly-cifar-tv-conv-v3-8.yaml
will-cromar: Done. b/154864921

PR #100, opened 2020-04-23
Move GCS bucket names to cluster-level ConfigMap.

will-cromar:
This makes dataset directories and model directories configurable at the cluster level. You should now be to run the same test config on any properly-configured cluster in any zone.

Also move `--model_dir` flag into test configs like we do with pytorch, since they're not used in every test.


k8s/us-central1/gcs-buckets.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
```
zcain117, line 3: so this gets pushed when we run `kubectl apply -f k8s/` and then other files can refer to it using:

```
              envFrom: [
                {
                  configMapRef: {
                    name: "gcs-buckets",
                  },
                },
              ],
```

Is that right? The reference is just based on `metadata.name` ? Or is there some other step to allow the other files to refer to this config yaml?

will-cromar: The config map is referenced by name. I deployed it manually this time to the right namespace (`--namespace=automated`), and I'll follow up with another change to update the postsubmit script.

PR #110, opened 2020-04-27
Add TPU pod tests for most supported TF models.

will-cromar:


templates/garden/nightly/bert-mnli.libsonnet
```diff
@@ -59,11 +59,29 @@ local tpus = import "../../tpus.libsonnet";
       '--eval_batch_size=32',
     ],
   },
+  local v2_32 = {
+    accelerator: tpus.v2_32,
+    command+: [
+      '--train_batch_size=128',
+      '--eval_batch_size=128',
```
allenwang28, line 66: BERT should be able to run with a batch size of 512

will-cromar: Fixed these batch sizes.

will-cromar:
Synced with @allenwang28. I will add metric bounds in a follow-up PR.

PR #113, opened 2020-04-28
Warn instead of alert for 404 jobs.

zcain117:
These alert emails are not actionable for the oncall. We should still look into why we have so many jobs in the pubsub queue that return 404 when we check for the job status.


taylanbil:
Thanks Zach.

PR #114, opened 2020-04-28
Rename TF model garden directories.

will-cromar:
Rename to `tensorflow` since we have non-model garden tests in there now.


zcain117:
> Can you do a quick pass over the READMEs as well?
> 
> I think at least this one refers to `garden` in its URLs: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/images/README.md

I think it's just that 1 readme actually:

https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/GoogleCloudPlatform/ml-testing-accelerators%24+garden+f:README.md&patternType=literal

will-cromar:
Good catch. Fixed.

PR #116, opened 2020-04-28
Add stable TF 2.2 tests.

will-cromar:


templates/tensorflow/r2.2/retinanet.libsonnet
```diff
@@ -0,0 +1,105 @@
...
+      "--tpu=$(KUBE_GOOGLE_CLOUD_TPU_ENDPOINTS)",
+      "--strategy_type=tpu",
+      "--params_override=%s" % (std.manifestYamlDoc(self.paramsOverride) + "\n"),
+      "--model_dir=$(MODEL_DIR)",
+    ],
+  },
+  local convergence = base.Convergence {
+    command+: [
+      "--mode=train",
+    ],
```
allenwang28, line 59: Hmm I suppose it would be non-trivial to add in the extra eval donut right?

templates/tensorflow/r2.2/transformer-translate.libsonnet
```diff
@@ -0,0 +1,78 @@
...
+local tpus = import "../../tpus.libsonnet";
+
+{
+  local transformer = base.GardenTest {
+    modelName: "transformer-translate",
+    command: [
+      "python3",
+      "official/nlp/transformer/transformer_main.py",
+      "--tpu=$(KUBE_GOOGLE_CLOUD_TPU_ENDPOINTS)",
+      "--steps_between_evals=10000",
```
allenwang28, line 26: If you want, you can set `steps_between_evals` to be the same as the `train_steps` so that the job runs faster

templates/tensorflow/r2.2/retinanet.libsonnet
will-cromar: It should be possible, but I haven't added good support for that case in the templates

templates/tensorflow/r2.2/transformer-translate.libsonnet
will-cromar: Good idea. Done.

PR #117, opened 2020-04-28
Don't retry oneshot tests by default.

### LINKED ISSUE #65: Oneshot retries may be unnecessary.
Idea: Oneshot runs are usually looked at and analyzed manually, so no automated restarts may be required in that case.

will-cromar:
Fix #65


taylanbil:
Thanks @will-cromar !

PR #131, opened 2020-05-05
Look up job memory usage in Google Cloud Monitoring and add to model …

zcain117:
…metrics.

I think the VM memory usage (i.e. `memory/used_bytes`) is useful both for us and for other users.

`accelerator/memory_used` may or may not be useful. I want to collect it for a few days from our models and see what it looks like.

### DIFF
--- a/metrics_handler/metrics_test.py
+++ b/metrics_handler/metrics_test.py
@@ -165,6 +165,18 @@ class MetricsTest(parameterized.TestCase):
     within_bounds = metrics.within_bounds(value, *bounds, inclusive)
     self.assertIs(within_bounds, expected)
 
+  # To use this test, replace the args to compute_memory_metrics with your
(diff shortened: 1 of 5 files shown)


metrics_handler/metrics_test.py
```diff
@@ -165,6 +165,18 @@ def test_within_bounds(self, value, bounds, inclusive, expected):
     within_bounds = metrics.within_bounds(value, *bounds, inclusive)
     self.assertIs(within_bounds, expected)
 
+  # To use this test, replace the args to compute_memory_metrics with your
```
will-cromar, line 168: If this can only be run manually, I suggest moving it to a different file/script.

zcain117: good idea, created a new file `manual_tests.py` for these and similar unit tests

PR #139, opened 2020-05-07
Add TPU health monitor container.

will-cromar:
This should warn us if the TPU leaves the HEALTHY state.

### DIFF
--- a/images/health-monitor/monitor.py
+++ b/images/health-monitor/monitor.py
@@ -29,7 +29,7 @@ flags.DEFINE_string('namespace', 'default', 'The namespace of the pod.')
 flags.DEFINE_string('container', 'train', 'The name of the container to watch.')
 
 flags.DEFINE_integer('interval', 60, 'Number of seconds to wait between health checks.')
-flags.DEFINE_bool('verbose', False, 'Whether to print when TPU is HEALTHY')
+flags.DEFINE_bool('verbose', False, 'Whether to print when TPU is HEALTHY.')
 
 flags.DEFINE_string('project', None, 'The GCP project with your GKE cluster.')
 flags.DEFINE_string('zone', None, 'The GCP zone with your GKE cluster.')
@@ -79,8 +79,8 @@ def main(_):
                     FLAGS.container, str(pod.status))
       exit(1)
     if getattr(status.state, 'terminated'):
-      logging.info('Container `%s` terminated with status:\n%s',
-                   FLAGS.container, str(status))
+      logging.warning('Container `%s` terminated with status:\n%s',
+                      FLAGS.container, str(status))
       break
   
     time.sleep(FLAGS.interval)
--- a/images/health-monitor/monitor.py
+++ b/images/health-monitor/monitor.py
@@ -1,0 +1,91 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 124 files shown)


images/health-monitor/monitor.py
```diff
@@ -0,0 +1,91 @@
...
+flags.DEFINE_bool('verbose', False, 'Whether to print when TPU is HEALTHY')
+
+flags.DEFINE_string('project', None, 'The GCP project with your GKE cluster.')
+flags.DEFINE_string('zone', None, 'The GCP zone with your GKE cluster.')
+
+def main(_):
+  if FLAGS.verbose:
+    logging.set_verbosity(logging.DEBUG)
+  else:
+    logging.set_verbosity(logging.WARNING)
```
zcain117, line 41: does this mean that `logging.info` messages would not appear in Stackdriver?

If so, just make sure you're OK with this log message below being at `info` level:

```
      logging.info('Container `%s` terminated with status:\n%s',
                   FLAGS.container, str(status))
```

images/health-monitor/monitor.py
```diff
@@ -0,0 +1,91 @@
...
+      logging.info('TPU health: %s', health)
+    else:
+      logging.warning('TPU health: %s', health)
+
+      if not tpu_client.recoverable():
+        logging.warning('TPU entered un-recoverable state: %s',
+                        tpu_client.state())
+        break
+
+    pod = k8s_client.read_namespaced_pod_status(FLAGS.pod, FLAGS.namespace)
```
zcain117, line 73: I've found with `read_namespaced_job_status` that I can and get a 404 not found for the job, but then query a few minutes or hours later and find the job. You might run into similar quirks for `pod_status`. 

I think it's fine to try this as-is but if we see flakiness we might need to be more forgiving of the API, e.g. trying a few times to get the status before giving up, rather than `exit(1)` on the first not found.

images/health-monitor/monitor.py
```diff
@@ -0,0 +1,91 @@
...
+from kubernetes import client, config
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string('pod', None, 'The name of the pod to monitor.')
+flags.DEFINE_string('namespace', 'default', 'The namespace of the pod.')
+flags.DEFINE_string('container', 'train', 'The name of the container to watch.')
+
+flags.DEFINE_integer('interval', 60, 'Number of seconds to wait between health checks.')
+flags.DEFINE_bool('verbose', False, 'Whether to print when TPU is HEALTHY')
```
zcain117, line 32: nit: end with period

zcain117:
Ah one more note: we should add some usage tips to our documentation. Maybe file for a bug for now. I feel like some kind of general playbook would be nice for people who have tests running. Include an entry for 'if your test crashes and stackdriver logs don't show errors, then check this other container and see what happened to TPU health'

images/health-monitor/monitor.py
will-cromar: That's a good point. I raised the log level because `cloud_tpu_client` produces quite a bit of log spam at the INFO level, and we didn't necessarily need to see that the TPU was healthy repeatedly. I think this last message could be useful, though. I'll go ahead and raise the log level of this to `warning`.

images/health-monitor/monitor.py
will-cromar: I tried running this a few times, and so far I haven't found a situation where the container is running, but the pod status can't be found. Let's see what happens here.

images/health-monitor/monitor.py
will-cromar: Done.

will-cromar:
Filed b/156025061.

I'll hold off on writing docs until we have a better idea of which of our random flakes are actually due to TPU issues.

zcain117:
> Filed b/156025061.
> 
> I'll hold off on writing docs until we have a better idea of which of our random flakes are actually due to TPU issues.

SGTM

PR #140, opened 2020-05-08
Metrics handler config recipes

zcain117:
- Update metrics handler README with examples of common config setups.
- Rename 1 field of the config and update the test yamls and `main.py` accordingly.
- Clean ups to the `examples/` yamls

### DIFF
--- a/metrics_handler/README.md
+++ b/metrics_handler/README.md
@@ -145,12 +145,145 @@ Full config options:
     "Accuracy/test_final"
   ],
 
-  # (Optional) Defaults to True. If set to False, this test will not write
-  # any errors to Stackdriver Error Reporting for out-of-bounds metrics.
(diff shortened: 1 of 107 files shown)


metrics_handler/README.md
```diff
@@ -145,12 +145,145 @@ Full config options:
...
 ```
+## Config "Recipes"
+
+See the `Config` section above for full details. Here are some common config patterns you might want to use.
+
+#### I want no alerts, no emails, and don't want to write to Bigquery.
+
+Best to turn off the metrics handler entirely. Delete the Cloud Function and the Cloud Scheduler to trigger the Cloud Function (both of which you added in the `Setup` section above).
+
+#### I want to record metrics in Bigquery but I don't alerts/emails.
```
will-cromar, line 165: nit: BigQuery

PR #146, opened 2020-05-11
Add tf-r2.2 tab and example test tab to dashboard.

zcain117:
Add an adjustment to the dashboard code for tabs that have very few tests (e.g. the examples tab) so that they can render in a reasonable way.


zcain117:
Deployed and dashboard LGTM

PR #148, opened 2020-05-12
Add in a few 1.15 tests

allenwang28:
This PR creates the tests for {ResNet, EfficientNet, MNasNet, Mask RCNN} x {v2-8, v3-8, v2-32, v3-32} on TF 1.15/TPUs.

Examples to verify functionality on v2-8s: http://shortn/_oCBd81CWIK


templates/tensorflow/r1.15/base.libsonnet
```diff
@@ -0,0 +1,14 @@
+local base = import "../base.libsonnet";
```
will-cromar, line 1: Remember to add a copyright header to new files.

templates/tensorflow/r1.15/base.libsonnet
```diff
@@ -0,0 +1,14 @@
...
+
+{
+  LegacyTpuTest:: base.LegacyTpuTest {
+    frameworkPrefix: "tf-r1.15",
+    tpuVersion: "1.15",
+    imageTag: "1.15",
+  },
+  Convergence:: mixins.Convergence {
+    # Run at 1:00 PST on Saturday
+    schedule: "0 8 * * 6"
```
will-cromar, line 12: Do we have to run these tests this frequently? Do we have to automate them at all, or can we just run them manually whenever we patch the 1.15 image?

I'm just asking -- I don't have a clear idea of what we need myself.

templates/tensorflow/r1.15/base.libsonnet
allenwang28: Done

templates/tensorflow/r1.15/base.libsonnet
allenwang28: Hmm, for now I don't think so. When 1.15.3 starts to roll out we might need to ramp it up - we'll probably have moved most BCT tests by then right?

PR #151, opened 2020-05-13
Create tutorial about basic templated one-shot tests.

will-cromar:
I had to do some hackery to make it so the monitor container generates for our tests but not the tutorial examples. Ideally, this should be an optional feature, and you should be able to easily specify the right image for your project. I'll file a larger bug about separating the reusable library components of our templates from the configuration that we use in our project.

PR #153, opened 2020-05-14
Remove misnamed 1.15 tests

allenwang28:


will-cromar:
I removed these in #151

PR #157, opened 2020-05-18
Log pubsub message ID in the metrics handler

zcain117:
TESTED=Ran local version of code against prod pubsub queue.

### DIFF
--- a/metrics_handler/main.py
+++ b/metrics_handler/main.py
@@ -545,9 +545,12 @@ def run_main(event, context):
   for msg in all_msgs:
     data_str = msg.message.data
     try:
+      message_id = msg.message.message_id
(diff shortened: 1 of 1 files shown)


metrics_handler/main.py
```diff
@@ -545,9 +545,12 @@ def run_main(event, context):
   for msg in all_msgs:
     data_str = msg.message.data
     try:
+      message_id = msg.message.message_id
+      logger.info('Found message_id: {}'.format(message_id))
       data = json.loads(data_str)
       data['publish_time'] = msg.message.publish_time.seconds
       data['ack_id'] = msg.ack_id
+      data['message_id'] = message_id
```
zcain117, line 553: `data` gets logged later in the function

PR #160, opened 2020-05-20
Add documentation on storage options.

will-cromar:


will-cromar:
I'm avoiding mentioning GCE PDs, since we had trouble managing those for our tests.

doc/storage.md
```diff
@@ -0,0 +1,52 @@
...
+  resources:
+    requests:
+      storage: 1Ki
+```
+
+`Job`s that rely on the dataset in Filestore will need to have the following line in their pod template to bind to that `PersistentVolumeClaim`:
+
+```yaml
+  ...
+  - name: filestore-datasets
```
zcain117, line 46: I'd add a comment on this line like "# `name` should match the `name` of your `PersistentVolume`"

(if that comment is accurate. Maybe the `name` on this line doesn't matter?)

doc/storage.md
```diff
@@ -0,0 +1,52 @@
...
+    persistentVolumeClaim:
+      claimName: filestore-datasets-claim
```
zcain117, line 48: comment here too, something like:  "# This should match the `name` of the PersistentVolumeClaim mentioned above."

doc/storage.md
```diff
@@ -0,0 +1,52 @@
...
+  ...
+```
```
zcain117, line 50: I'd add a line after this config snippet on how to refer to the dataset. In the test command that the user writes, what is the path to access the data?

doc/storage.md
will-cromar: `name` doesn't matter here. I added a note above `claimName`.

doc/storage.md
will-cromar: Done.

doc/storage.md
will-cromar: Done.

PR #165, opened 2020-05-21
Add script to find peak times of TPU usage.

zcain117:


zcain117:
Example output
Screen Shot 2020-05-20 at 12 55 08 PM

scripts/find_busy_times.py
```diff
@@ -0,0 +1,130 @@
...
+import croniter
+
+from absl import app
+from absl import flags
+
+FLAGS = flags.FLAGS
+flags.DEFINE_string('dir_path', None, 'Dir to search for test configs.')
+flags.DEFINE_integer('chunk_minutes', 10, 'Number of minutes per chunk.')
+
+_DEADLINE_REGEX = r'"?activeDeadlineSeconds"?: "?(\d+)"?'
```
will-cromar, line 49: FYI, this might not work nicely if we set a deadline on individual pods in addition to the job deadline. You may be better off parsing the files with pyyaml and getting `spec.template.spec.activeDeadlineSeconds` directly. I'll leave that up to you.

PR #167, opened 2020-05-27
Metric comparison

zcain117:
Add new page to the dashboard to compare metric(s) across multiple tests.

TESTED=unit testing and ran local bokeh server to try out the new page.


zcain117:
Right now there's 1 plot per metric and 1 line per test on each plot. I added a color-coded table of the same data beneath each plot to give some flexibility and can adapt more based on whatever comes out as the main usage patterns

Screen Shot 2020-05-27 at 3 50 01 PM

PR #168, opened 2020-05-27
Update TF 1.15 image to 1.15.3

will-cromar:


images/tensorflow-tpu-1x/cloudbuild.yaml
```diff
@@ -25,6 +25,6 @@ steps:
 images: ['gcr.io/$PROJECT_ID/tensorflow-tpu-1x:$_VERSION']
 timeout: 1200s  # 20 minutes
 substitutions:
-  _VERSION: 'latest'
+  _VERSION: 'r1.15.3'
```
zcain117, line 28: seems theres no trigger for this image so I'm assuming you'll run this manually

PR #173, opened 2020-05-29
Wait longer before giving up for missing k8s jobs.

zcain117:
context: https://b.corp.google.com/issues/157749796

Kubernetes sometimes returns 404 for jobs temporarily. Extend the timeout before we decide a job is truly missing.


metrics_handler/main.py
```diff
@@ -42,6 +42,7 @@
 BQ_METRIC_TABLE_NAME = 'metric_history'
 METRICS_WRITTEN_TOPIC = 'metrics-written'
 MIN_MSG_AGE_SEC = 60
+MAX_SEC_TO_WAIT_FOR_MESSAGE = 48 * 60 * 60
```
will-cromar, line 45: nit: Can you add a comment with the conversion to hours?

zcain117: Done

PR #183, opened 2020-06-03
Edit r1.15 transformer command so pods initialize correctly.

taylanbil:
`--cloud_tpu_name` was leading to an [error](b/158014511):

```
INFO:tensorflow:Querying Tensorflow master (grpc://10.120.19.18:8470,grpc://10.120.19.20:8470,grpc://10.120.19.21:8470,grpc://10.120.19.19:8470) for TPU system metadata.
Error recorded from training_loop: Could not interpret "10.120.19.18:8470,grpc://10.120.19.20:8470,grpc://10.120.19.21:8470,grpc://10.120.19.19:8470" as a host-port pair.
```
Replaced w/ `--tpu`

I've been trying for 2 days now to get a pod run to verify the fix, but unfortunately I can't seem to get available tpu pods for it. In the meantime, getting this out there to collect comments.

failed oneshots because no capacity:
[oneshot for v2-32 func](https://pantheon.corp.google.com/kubernetes/job/us-central1-b/oneshots/default/tf-r1.15-transformer-func-v2-32-4p7sb/details?project=xl-ml-test)
[oneshot for v3-32 func](https://pantheon.corp.google.com/kubernetes/job/us-central1-b/oneshots/default/tf-r1.15-transformer-func-v3-32-qv5zk/details?project=xl-ml-test)
success: 
[oneshot for v2-8 func](https://pantheon.corp.google.com/kubernetes/job/us-central1-b/oneshots/default/tf-r1.15-transformer-func-v2-8-tbm9g/details?project=xl-ml-test)

### DIFF
--- a/k8s/europe-west4/gen/tf-r1.15-transformer-conv-v2-32.yaml
+++ b/k8s/europe-west4/gen/tf-r1.15-transformer-conv-v2-32.yaml
@@ -49,7 +49,7 @@
             - "--use_tpu=True"
             - "--schedule=train"
             - "--data_dir=$(T2T_TRANSFORMER_DIR)"
-            - "--cloud_tpu_name=$(KUBE_GOOGLE_CLOUD_TPU_ENDPOINTS)"
+            - "--tpu=$(KUBE_GOOGLE_CLOUD_TPU_ENDPOINTS)"
             - "--output_dir=$(MODEL_DIR)"
             - "--iterations_per_loop=5000"
             - "--tpu_num_shards=32"
--- a/k8s/europe-west4/gen/tf-r1.15-transformer-conv-v3-32.yaml
+++ b/k8s/europe-west4/gen/tf-r1.15-transformer-conv-v3-32.yaml
@@ -49,7 +49,7 @@
             - "--use_tpu=True"
             - "--schedule=train"
             - "--data_dir=$(T2T_TRANSFORMER_DIR)"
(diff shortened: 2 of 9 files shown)


zcain117:
Looks promising to me but @will-cromar is the TF pods expert

will-cromar:
`t2t-trainer` doesn't actually have a `--tpu` flag registered, as near as I can tell. If this works, I suspect that it may be getting the TPU address from the environment, and you would get the same effect if you removed the `--cloud_tpu_name` flag altogether. 

Those pod runs didn't schedule because you were running on the cluster in `us-central1-b`. If you connect to `europe-oneshots` instead, you should be able to schedule pod jobs. `gcloud container clusters get-credentials europe-oneshots --zone europe-west4-a --project xl-ml-test`

taylanbil:
So what's the best way to pass this to t2t-trainer? Getting from the environment sounds unreliable to me.

will-cromar:
I'm surprised that passing `$KUBE_GOOGLE_CLOUD_TPU_ENDPOINTS` works for donuts but not pods. For our BCT tests, we usually passed in the GCE VM's `$TPU_NAME`, but for some reason GKE has its own format for `$TPU_NAME` (`zone/name`). The only option I can think of here is to use `utils.scriptCommand` or  add something to the init entrypoint to reformat the `$TPU_NAME` variable to work with TF. You can assign that work to me if you want to.

taylanbil:
Ok, closing this PR then. I'll assign you the b/. Thanks.

PR #185, opened 2020-06-03
Fix metric collection for TF tests.

will-cromar:
- Distinguish metrics in different runs.
- Update regression config for classifier-* tests.
- Add support for averaging metrics.
- Assert on the average metric value for examples/sec in TF nightly tests.

Tested manually in test project.

### DIFF
--- a/metrics_handler/README.md
+++ b/metrics_handler/README.md
@@ -60,6 +60,7 @@ Full config options:
   # "final": Save the final value for that metric.
   # "max": Save the max value for that metric.
   # "min": Save the min value for that metric.
+  # "average": Save the average value for that metric.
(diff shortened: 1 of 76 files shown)


metrics_handler/README.md
```diff
@@ -85,7 +85,12 @@ Full config options:
   },
 
   # (Optional) Defaults to True. Set to false to disable all BigQuery writes.
-  "write_to_bigquery": "True"
+  "write_to_bigquery": "True",
```
zcain117, line 88: Up above in this file it lists 'Available strategies:'
Could you add "average" to that list?

will-cromar: Done.

PR #192, opened 2020-06-08
Use fewer steps for TF transformer GPU tests.

zcain117:


tests/tensorflow/nightly/transformer-translate.libsonnet
```diff
@@ -39,6 +39,11 @@ local gpus = import "templates/gpus.libsonnet";
       "--train_steps=20000",
     ],
   },
+  local functional_10k_steps = mixins.Functional {
```
will-cromar, line 42: nit: Can you make this called `functional_short` or something like that?

zcain117: done

PR #198, opened 2020-06-15
Double the timeout for EfficientNet donut convergence tests

allenwang28:
It seems to time out at around 300 epochs in 30 hours. Doubling the timeout should allow it to finish.


zcain117:
can we reduce num epochs?  2.5 days between code change and results might make things hard to debug

allenwang28:
Good point, I reduced the epochs to 350 which should still surpass the 0.76 number (for reference, 500 epochs is required to get to the original EfficientNet accuracy)

allenwang28:
I added 5 more hours just to be safe

PR #199, opened 2020-06-16
Update README to link to tests instead of templates

allenwang28:
Not sure if we want to still link to templates but the README does not exist there anymore.


will-cromar:
That should still link to templates/. Can you change the link to that directory and I'll add a readme?

allenwang28:
Adding a README to templates sounds better, will close this PR instead

PR #202, opened 2020-06-16
Add automated script to clean up OS Login keys.

will-cromar:
Fix bug where unnecessary quotas and namespaces were created on oneshot clusters.

### DIFF
--- a/k8s/europe-west4/profile-cleanup.yaml
+++ b/k8s/europe-west4/profile-cleanup.yaml
@@ -1,0 +1,36 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: profile-cleanup
+spec:
+  schedule: "0 0 * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: deleter
+            image: google/cloud-sdk
+            args:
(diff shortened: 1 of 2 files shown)


k8s/europe-west4/profile-cleanup.yaml
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: profile-cleanup
+spec:
+  schedule: "0 0 * * *"
```
jysohn23, line 20: Any chance this could cause problem if run concurrently with a pods test? I'm assuming its fairly unlikely?

will-cromar: I think it's pretty unlikely, but I'll keep an eye out for issues.

postsubmit.yaml
```diff
@@ -39,12 +39,12 @@ steps:
...
   - 'CLOUDSDK_CONTAINER_CLUSTER=xl-ml-test-europe-west4'
 - name: 'gcr.io/cloud-builders/kubectl'
-  args: ['apply', '-f', 'k8s/us-central1/', '-f', 'k8s/common/',]
+  args: ['apply', '-f', 'k8s/us-central1/gcs-buckets.yaml', '-f', 'k8s/us-central1/volumes.yaml', '-f', 'k8s/common/',]
   env:
   - 'CLOUDSDK_COMPUTE_ZONE=us-central1-b'
   - 'CLOUDSDK_CONTAINER_CLUSTER=oneshots'
 - name: 'gcr.io/cloud-builders/kubectl'
-  args: ['apply', '-f', 'k8s/europe-west4/', '-f', 'k8s/common/',]
+  args: ['apply', '-f', 'k8s/europe-west4/gcs-buckets.yaml', '-f', 'k8s/common/',]
```
zcain117, line 47: I'm confused here and on line 42: are these changes related to the new `profile-cleanup.yaml` file? It's fine to add these changes even if they're unrelated. Just wanted to make sure you didn't mean to add `profile-cleanup.yaml` to the postsubmit script

will-cromar: It's related. The idea is to keep the profile cleanup job from being created once in the main cluster and once on the oneshot cluster. I'm specifying only the files that need to be created on oneshots.

PR #209, opened 2020-06-23
Make linking `gcs-buckets` configmap optional.

will-cromar:


will-cromar:
I left the default the same for now, so this is a no-op for existing tests.

PR #212, opened 2020-06-23
Refactor accelerator mixins and preemptible TPU logic.

will-cromar:
- Change `accelerator.PodSpec` to `accelerator.PodTemplate(tpuSettings)` that encapsulates all TPU logic. Remove hacky `if type == 'tpu'` conditionals from base template.
- Create explicit `BaseAccelerator` base object.
- Fixes bug/inconsitency where `model + tpu + mode` gave a different result than `model + mode + tpu`
- Remove empty objects from GPU job metadata.
- Replace `tpus.Preemptible` with `mixins.PreemptibleTpu`.
- Fix bug where we were setting a no-op `preemptible` field on GPU accelerators.

### DIFF
--- a/templates/tpus.libsonnet
+++ b/templates/tpus.libsonnet
@@ -12,8 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+local base = import 'base.libsonnet';
+
 {
-  TpuSpec:: {
+  TpuSpec:: base.BaseAccelerator {
     local tpu = self,
 
     name: "v%(version)d-%(size)d" % tpu,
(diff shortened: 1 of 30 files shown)


templates/tpus.libsonnet
```diff
@@ -23,24 +25,34 @@
...
+                else
+                  "",
+              local resourceName = "cloud-tpus.google.com/%sv%s" % [preemptiblePrefix, tpu.version],
+
+              limits+: { [resourceName]: tpu.size },
+            },
           },
         },
+        nodeSelector+:
+          if tpuSettings.requireTpuAvailableLabel then
```
zcain117, line 50: Do you need to set this to True somewhere? Looks like default is False in the base template

tests/pytorch/nightly/resnet50-mp.libsonnet
```diff
@@ -75,7 +75,7 @@ local tpus = import "templates/tpus.libsonnet";
     accelerator: tpus.v3_8,
   },
   configs: [
-    resnet50_MP + v3_8 + convergence + timeouts.Hours(26),
+    resnet50_MP + v3_8 + convergence + timeouts.Hours(26) + mixins.PreemptibleTpu,
```
zcain117, line 78: I don't see this changed .yaml file in this PR. Using preemptible here could make sense to collect data but I feel like such a long running test is going to get preempted a lot

templates/tpus.libsonnet
will-cromar: This is set to true in `tests/common.libsonnet`. This selector only makes sense in the context of a regional cluster, which is going to be overkill for most teams, so the default is false.

tests/pytorch/nightly/resnet50-mp.libsonnet
will-cromar: This test was already using preemptible. See the change on line 58.

PR #220, opened 2020-06-29
Fix multi-k80 tests on TF nightly.

will-cromar:
Tested each of these for a few minutes to make sure they didn't freeze up as before. I'll keep an eye out for potential performance regressions.


tests/tensorflow/nightly/transformer-translate.libsonnet
```diff
@@ -70,6 +70,9 @@ local gpus = import "templates/gpus.libsonnet";
   },
   local k80x8 = k80 {
     accelerator: gpus.teslaK80 + { count: 8 },
+    command+: [
+      "--all_reduce_alg=hierarchical_copy",
```
zcain117, line 74: This looks like it took a while to find; sounds obscure
Good find though!

PR #221, opened 2020-06-30
Add TensorFlow-2.3 TPU tests.

aman2930:


tests/tensorflow/r2.3/bert-mnli.libsonnet
```diff
@@ -0,0 +1,86 @@
...
+  },
+  local convergence = common.Convergence {
+    command+: [
+      '--num_train_epochs=3',
+    ],
+  },
+  local v2_8 = {
+    accelerator: tpus.v2_8,
+    command+: [
+      '--train_batch_size=128',
```
will-cromar, line 50: Did you remember to update your master branch before copying these files? It looks like you're missing recent changes like the change to this batch size

will-cromar: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/tensorflow/nightly/bert-mnli.libsonnet#L51

PR #229, opened 2020-07-14
ShapeMask convergence and evaluation tests

gagika:
Adding ShapeMask convergence and evaluation test.
Tested a couple of functional and convergence tests on a donut and pod.

### DIFF
--- a/tests/tensorflow/r2.3/shapemask.libsonnet
+++ b/tests/tensorflow/r2.3/shapemask.libsonnet
@@ -50,36 +50,36 @@ local utils = import "templates/utils.libsonnet";
         shape_prior_path: "gs://cloud-tpu-checkpoints/shapemask/kmeans_class_priors_91x20x32x32.npy",
       },
     },
-    trainCommand(params_override)::
-      utils.scriptCommand(
-        |||
-          %(common)s --mode=train \
-          --params_override="%(params_override)s"
-        ||| % {common: command_common, params_override: std.manifestYamlDoc(params_override)}
-      ),
-    trainAndEvalCommand(params_override)::
-      utils.scriptCommand(
-        |||
-          %(common)s --mode=train \
-          --params_override="%(params_override)s"
-          %(common)s --mode=eval \
-          --params_override="%(params_override)s"
-        ||| % {common: command_common, params_override: std.manifestYamlDoc(params_override)}
-      ),
   },
   local functional = mixins.Functional {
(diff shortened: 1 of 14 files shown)


tests/tensorflow/r2.3/shapemask.libsonnet
```diff
@@ -107,18 +114,36 @@ local tpus = import "templates/tpus.libsonnet";
       train+: {
         batch_size: 256,
       },
+      eval+: {
+        batch_size: 64,
+      },
     },
   },
-
+  local shapemask_func_v2_8 = shapemask + functional + v2_8 {
```
will-cromar, line 122: I'd format this a little differently. Since the JSonnet document is lazily evaluated, I think you could have something like this:

```
local convergence = {
  command: utils.scriptCommand(
       |||
          %(common)s --mode=train \
          --params_override="%(params_override)s"
          %(common)s --mode=eval \
          --params_override="%(params_override)s"
       ||| % {common: command_common, params_override: std.manifestYamlDoc(params_override)}
}
```

And similar for `functional`, then we could keep the pattern of `shapemask + functional + v2_8`. Would that work?

gagika: I tried your suggestion,  I could access params_override in the function call was through `convergence.params_override` (`self.params_override` or `params_override` error out), but then it just picks up `paramsOverride` within convergence (not lazily) and doesn't pick up from shapemask when using `shapemask + convergence + v3_32` in config:

```
local convergence = mixins.Convergence {
    paramsOverride+: {
      train+: {
        total_steps: 22500,
      },
    },
    command: utils.scriptCommand(
        |||
            %(common)s --mode=train \
            --params_override="%(params_override)s"
            %(common)s --mode=eval \
            --params_override="%(params_override)s"
        ||| % {common: command_common, params_override: std.manifestYamlDoc(convergence.paramsOverride)}
    ),
  },
```

will-cromar: Oh, I see. You'll have to add `local config = self` in `convergence` and reference the overrides as `config.paramsOverride`. See this for example: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/tensorflow/nightly/retinanet.libsonnet#L82

gagika: Thanks, Done

PR #230, opened 2020-07-14
Add PT XLA resnet50 TPU pod test that uses real imagenet

will-cromar:
- Move resnet50 pods config to a new file, since the pattern is different.
- Add volume support in `launch_k8s_workers.py`
- Add `volumes.yaml` config for `europe-west4` cluster. (Created Filestore instance in our project.)
- Remove `-claim` suffix from volume base object. Add suffix to PyTorch dataset volume. (no-op naming change for our existing tests.)

I tested this with a different NFS server setup. I'm not entirely sure that Filestore will be performant enough for all 4 client workers to drive input. If not, I'll follow up with another config change.


images/pytorch-pods/launch_k8s_workers.py
```diff
@@ -148,6 +152,15 @@ def main(argv):
     'TPU_NUM_DEVICES': '8',
   }
 
+  if FLAGS.volumes:
+    volumes = {
+      name: mount_path for name, mount_path in
+      [v.split(':') for v in FLAGS.volumes]
+    }
+    print(volumes)
```
zcain117, line 160: In some places you use `logging` but here `print`. Just confirm this is what you wanted

images/pytorch-pods/launch_k8s_workers.py
```diff
@@ -148,6 +152,15 @@ def main(argv):
...
+  else:
+    volumes = []
```
zcain117, line 162: if populated, it's a dict but if empty, it's a list. Probably fine since `if not volumes` will work the same but there might be some edge cases that lead to weird behavior

images/pytorch-pods/launch_k8s_workers.py
```diff
@@ -190,8 +203,32 @@ def main(argv):
...
+        'volumes': [
+          {
+            'name': 'dshm',
+            'emptyDir': { 'medium': 'Memory' },
+          },
+          *[
+            {
+              'name': name,
+              'persistentVolumeClaim': { 'claimName': name }
+            } for name, _ in volumes.items()
```
zcain117, line 229: Oh here would be probably be a bad spot to have that empty list mentioned in my other comment

tests/pytorch/nightly/resnet50-pod.libsonnet
```diff
@@ -0,0 +1,72 @@
...
+local tpus = import "templates/tpus.libsonnet";
+local mixins = import "templates/mixins.libsonnet";
+
+{
+  local resnet50_pod = common.PyTorchGkePodTest {
+    modelName: "resnet50-mp",
+    command: [
+      "python3",
+      "/pytorch/xla/test/test_train_mp_imagenet.py",
+      # "--logdir=$(MODEL_DIR)",
```
zcain117, line 26: intended to be left commented?

tests/pytorch/nightly/resnet50-pod.libsonnet
```diff
@@ -0,0 +1,72 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+local common = import "common.libsonnet";
+local timeouts = import "templates/timeouts.libsonnet";
+local tpus = import "templates/tpus.libsonnet";
+local mixins = import "templates/mixins.libsonnet";
+
+{
+  local resnet50_pod = common.PyTorchGkePodTest {
```
zcain117, line 21: is this using some default timeout?

images/pytorch-pods/launch_k8s_workers.py
will-cromar: Oops, that's left over form debugging.

images/pytorch-pods/launch_k8s_workers.py
will-cromar: Fixed.

images/pytorch-pods/launch_k8s_workers.py
will-cromar: Fixed.

tests/pytorch/nightly/resnet50-pod.libsonnet
```diff
@@ -0,0 +1,71 @@
...
+  },
+  local convergence = common.Convergence {
+    command+: [
+      "--num_epochs=90",
+      "--datadir=/datasets/imagenet",
+      "--batch_size=128",
+      "--log_steps=200",
+    ],
+
+    workerCpu: "90",
```
jysohn23, line 44: Are we reducing this per our discussion?

will-cromar: Done. Lowered it to 36 CPU/100Gi, which is still above the maximum utilization that I'm seeing.

PR #238, opened 2020-07-23
Remove Tensorboard logging from Fairseq Transformer test.

zcain117:
This logging was causing the test to freeze. Filed a bug to re-enable the logging


will-cromar:
Remember to run `gen-tests.sh`

taylanbil:
Thanks Zach.

PR #243, opened 2020-07-28
Disable Tensorboard for Pytorch Roberta pretraining test.

zcain117:
Was seeing the same issue for this fairseq roberta test that I saw with fairseq transformer: https://b.corp.google.com/issues/162248420

Training just gets stuck at some point, though the amount of time until getting stuck is variable.

I have a bug to re-enable Tensorboard later but for now neither this nor the transformer test was using any alerts based on Tensorboard metrics


zcain117:
@taylanbil FYI

taylanbil:
Yeah this is a fairseq thing rather than roberta/transformer. We need to
figure out why we cant use tensorboardx at some point.

PR #245, opened 2020-07-29
Use upgraded pillow version

zcain117:


zcain117:
I pushed a new deployment of the dashboard with this change and it's working fine AFAICT

PR #252, opened 2020-08-03
Sanitize job name before launching jobs in launch_k8s_workers.py

zcain117:
TESTED=Simple local testing.

Let me know if you think it's worth adding a test file.

### DIFF
--- a/images/pytorch-pods/launch_k8s_workers.py
+++ b/images/pytorch-pods/launch_k8s_workers.py
@@ -75,6 +75,18 @@ flags.DEFINE_list('volumes', None,
 def _format_env(envs):
   return [{'name': k, 'value': v} for k, v in envs.items()]
 
+# Name must consist of lower case alphanumeric characters or '-'.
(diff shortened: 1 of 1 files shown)


images/pytorch-pods/launch_k8s_workers.py
```diff
@@ -75,6 +75,18 @@
 def _format_env(envs):
   return [{'name': k, 'value': v} for k, v in envs.items()]
 
+# Name must consist of lower case alphanumeric characters or '-'.
+def _sanitize_job_name(name):
```
will-cromar, line 79: IMO it would be better to use `re.sub` or similar than to imperatively build a new string. ie something like `re.sub(r'[^a-z0-9\-]', '-',  name.lower())`

zcain117: SGTM

zcain117: tested again locally

PR #255, opened 2020-08-04
Add PerfZero benchmarks from TF model garden

will-cromar:
- Add PerfZero to TF image.
- Add v100-x8 cluster to support existing multi-GPU benchmarks.

I'm using a temporary hack to replace some hard-coded paths in the benchmark using `sed`. I did test runs of almost all of these configs and confirmed that the job status (ie exit code) does show up on our dashboard.


tests/tensorflow/nightly/perfzero.libsonnet
```diff
@@ -0,0 +1,275 @@
...
+local tpus = import "templates/tpus.libsonnet";
+local gpus = import "templates/gpus.libsonnet";
+local utils = import "templates/utils.libsonnet";
+
+{
+  local PerfZeroTest = common.ModelGardenTest {
+    local config = self,
+
+    timeout: timeouts.one_hour,
+    schedule: "0 8 * * *",
```
zcain117, line 12: Just confirming this is a good time to catch the new nightly builds. It's the same time we use for the default `func` tests: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/templates/mixins.libsonnet#L22

If you see flakiness due to resource shortages, could be worth trying to stagger and see if it has any effect.

tests/tensorflow/nightly/perfzero.libsonnet
```diff
@@ -0,0 +1,275 @@
...
+      },
+      table: {
+        project: "xl-ml-test",
+        name: "perfzero_dataset.perfzero_table",
+      },
+    },
+    command: [
+      "python3",
+      "/benchmarks/perfzero/lib/benchmark.py",
+      "--gcloud_key_file=",
```
zcain117, line 29: making sure this is intended and not that the arg was redacted by Github or something

will-cromar: Yes, this need to be an empty string to indicate that we should use the environment ADC instead of a key file.

tests/tensorflow/nightly/perfzero.libsonnet
will-cromar: This is what we use for our current TensorFlow suite, but these tests should be much shorter. I will adjust timing if necessary.

tests/tensorflow/nightly/perfzero.libsonnet
allenwang28: So just to clarify, this means that this runner doesn't require the gcloud key hack that the MLCompass runner does?

will-cromar: That's correct. The test runs in GKE, which provides application default credentials instead: https://cloud.google.com/docs/authentication/production

will-cromar:
Not at the moment. We'll probably need to do work on the MLCompass runner to better enable the internal use case first.  This is a first step that will let us share comparable performance data and share the burden of maintaining the tests that run most frequently (which is most important IMO).

PR #256, opened 2020-08-05
Add test that TPU memory gets reset between training runs.

zcain117:
Requested in our PRD


zcain117:
forgot to link the oneshot,   but it passed

PR #262, opened 2020-08-12
Remove examples tab from dashboard.

zcain117:
I will deploy a new revision of the dashboard once this PR goes in.


zcain117:
Deploying new version of dashboard now with this change

PR #264, opened 2020-08-12
Scrape metrics from perfzero summary JSON when available.

will-cromar:
Give preference to PerfZero metrics when available, including for "computed" values like wall time. No-op for non-perfzero tests.

### DIFF
--- a/metrics_handler/main.py
+++ b/metrics_handler/main.py
@@ -19,6 +19,7 @@ import io
 import itertools
 import json
 import math
+import os
 import time
 import traceback
 import uuid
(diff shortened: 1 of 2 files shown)


metrics_handler/main.py
```diff
@@ -499,12 +533,18 @@ def _process_pubsub_message(msg, status_handler, logger):
         'job_status was `{}` for test `{}`'.format(
             job_status['final_status'], test_name),
         debug_info=debug_info)
-
+
```
zcain117, line 536: trailing space

PR #266, opened 2020-08-13
Delete remaining TF r2.3 tests.

will-cromar:


will-cromar:
I'm not actually deleting these tests entirely. I'm making the default schedule for everything except resnet `null` so that they don't run automatically. If we patch r2.3, we'll need to either re-enable the tests. You can still run them as oneshots as well.

PR #268, opened 2020-08-14
Add required_metrics option to metrics handler.

zcain117:
Useful feature for upstream models whose metric names might change.

### DIFF
--- a/tests/pytorch/nightly/hf-glue.libsonnet
+++ b/tests/pytorch/nightly/hf-glue.libsonnet
@@ -55,7 +55,6 @@ local utils = import "templates/utils.libsonnet";
       ||| % {common: command_common, common_copy: command_copy_metrics}
     ),
     regressionTestConfig+: {
-      alert_for_failed_jobs: false,
       metric_success_conditions+: {
         "eval_mnli/acc": {
           success_threshold: {
@@ -83,7 +82,6 @@ local utils = import "templates/utils.libsonnet";
       ||| % {common: command_common, common_copy: command_copy_metrics}
     ),
     regressionTestConfig+: {
-      alert_for_failed_jobs: false,
       metric_success_conditions+: {
         "eval_mnli/acc": {
           success_threshold: {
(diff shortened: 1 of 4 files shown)


tests/pytorch/nightly/hf-glue.libsonnet
```diff
@@ -167,7 +163,7 @@ local utils = import "templates/utils.libsonnet";
       ||| % {common: command_common, common_copy: command_copy_metrics}
     ),
     regressionTestConfig+: {
-      alert_for_failed_jobs: false,
+      required_metrics: ['eval_mnli/acc'],
```
zcain117, line 166: @jysohn23 let me know which metrics are good to include here

zcain117:
I'll finalize the `required_metrics` for 1.6 and nightly test files and then generate new test configs once I know which required metrics to use

tests/pytorch/nightly/hf-glue.libsonnet
jysohn23: Let's not include these yet since those metrics haven't been renamed yet and I'm still looking at a fix for the regression. Let's add then once I get to HF fix.

tests/pytorch/nightly/hf-glue.libsonnet
```diff
@@ -167,7 +163,7 @@ local utils = import "templates/utils.libsonnet";
       ||| % {common: command_common, common_copy: command_copy_metrics}
     ),
     regressionTestConfig+: {
-      alert_for_failed_jobs: false,
```
jysohn23, line 170: For my understanding, what does `alert_for_failed_jobs` do?

zcain117: if the GKE job ends in failure status, the metrics handler will normally send an alert. But if this setting is turned to `false`, then it won't trigger an alert. OOB metrics would still trigger alerts

https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/metrics_handler/main.py#L529

We added this way back when you said that the tests were unstable. Note that the tests still show as failures in the dashboard, they just won't send email alerts

zcain117:
Reverted the changes to HF test file @jysohn23 
I can make a follow-up PR with changes to the `nightly` and `r1.6` HF test files to add the `required_metrics` and remove the `alert_for_failed_jobs: false,` lines

PR #272, opened 2020-08-26
Add DLRM convergence tests on Criteo Kaggle dataset

jysohn23:
Keras API CTL test changes come from just running `scripts/gen-tests.sh`.


zcain117:
nice job adding this!
Good to have an example of parsing logs in case anyone has a model that doesn't write to Tensorboard

jysohn23:
I still think it's better to log metric to Tensorboard since your framework tracks for regressions but I think we can do that later 😄

PR #273, opened 2020-08-26
Bisection tool

zcain117:
Add an example of simple regression over Docker images. Useful for pytorch TPU tests since we have an hourly docker image that can be used with this setup to find regression points.

Ran locally with a few combinations of passing/failing `command`s


bisection/run.sh
```diff
@@ -0,0 +1,105 @@
+# NOTE: This algorithm assumes that the test you supply follows the pattern of:
```
will-cromar, line 1: Add copyright header

bisection/bisection_template.jsonnet
```diff
@@ -0,0 +1,39 @@
+local base = import 'templates/base.libsonnet';
```
will-cromar, line 1: Copyright header

zcain117:
things got weird with binary files when I ran the addlicense command

copied the relevant files (with licenses added) to https://github.com/GoogleCloudPlatform/ml-testing-accelerators/pull/276

PR #275, opened 2020-09-01
Add RoBERTa base MLM pre-training and finetuning tasks

jysohn23:
Also updates per_gpu to be per_device as we're now deprecating the
per_gpu flag on HF size to be more TPU friendly.

### DIFF
--- a/tests/pytorch/nightly/hf-glue.libsonnet
+++ b/tests/pytorch/nightly/hf-glue.libsonnet
@@ -55,7 +55,7 @@ local utils = import "templates/utils.libsonnet";
       ||| % {common: command_common, common_copy: command_copy_metrics}
     ),
     regressionTestConfig+: {
-      alert_for_failed_jobs: false,
+      required_metrics: ['eval_mnli/acc', 'eval_mnli-mm/acc'],
       metric_success_conditions+: {
         "eval_mnli/acc": {
           success_threshold: {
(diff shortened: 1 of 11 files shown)


tests/pytorch/nightly/hf-glue.libsonnet
```diff
@@ -105,8 +105,8 @@ local utils = import "templates/utils.libsonnet";
     command: utils.scriptCommand(
       |||
         %(common)s --model_name_or_path roberta-large \
-        --per_gpu_train_batch_size 16 \
-        --per_gpu_eval_batch_size 16
+        --per_device_train_batch_size 16 \
+        --per_device_eval_batch_size 16
```
zcain117, line 109: For these tests, you have metrics assertion for `eval_mnli/acc` and `eval_mnli-mm/acc`
if you think those keys will be pretty stable, I'd recommend adding:
`required_metrics: ['eval_mnli/acc', 'eval_mnli-mm/acc']`
to each of the remaining tests in this file

jysohn23:
> Is there any test for the new modeling work that we might also add to their CI? Right now they're running on a timed basis, not every PR, so I think we could run a long-ish test (maybe 10-20 min?)
> This is not blocking this PR, just wondering

On that we could add a quick fine-tuning task for, let's say RoBERTa, on a smaller LM dataset like wikitext2, which should be pretty quick.

tests/pytorch/nightly/hf-glue.libsonnet
jysohn23: Done.

PR #283, opened 2020-09-11
Use only past 30 days of data for heatmap and metrics dashboards.

zcain117:
Tested by running the dashboard locally. It feels significantly faster with this change.


will-cromar (approved):

zcain117:
Deployed a new version of the dashboard with the changes and it looks good

PR #290, opened 2020-09-23
Add step in tutorial to connect to GKE cluster.

will-cromar:


sshahrokhi (approved):

zcain117 (approved):
maybe add this command to the [other tutorial](https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/doc/tutorials/test-templates-automated.md) as well. The cluster create command is there in the other tutorial so presumably the user would also have to install GPU drivers again

PR #291, opened 2020-09-23
a few v100 resnet tests

sshahrokhi:
a few tests to be shown on the dashboard (resnet on v100, functional and convergence)


tests/tensorflow/nightly/classifier-resnet.libsonnet
```diff
@@ -146,7 +146,8 @@ local gpus = import "templates/gpus.libsonnet";
     resnet + v100 + functional + timeouts.Hours(3) + mixins.Suspended,
     resnet + v100x4 + functional + mixins.Experimental,
     resnet + v100x4 + convergence + mixins.Experimental,
-    resnet + v100x8 + functional + mixins.Experimental,
+    resnet + v100x8 + functional,
+    resnet + v100x8 + convergence,
```
will-cromar, line 150: The default timeout for convergence tests is 10 hours. @allenwang28 Do you know the runtime for classifier-resnet on 8-v100s?

will-cromar (approved):
LGTM!

tests/tensorflow/nightly/classifier-resnet.libsonnet
allenwang28: Hmmm, good point - 6 months ago the answer was around [2.5 days](https://mldash.corp.google.com/experiments/8637395369934579383#scalars).

I think we intend to tear this test down so it's fine if it fails!

sshahrokhi: yes thanks Allen, I was about to tell that, I don't think it matters if it fails! Thank you both for your helps!

PR #295, opened 2020-09-29
XLA GPU tests for PyTorch

zcain117:
TESTED=oneshots for mnist and `kubectl apply -f` for resnet and the resnet workload passed this morning


tests/pytorch/nightly/resnet50-mp.libsonnet
```diff
@@ -13,11 +13,48 @@
...
+      datasets: common.datasetsVolume
+    },
+    jobSpec+:: {
+      template+: {
+        spec+: {
+          containerMap+: {
+            train+: {
+              resources+: {
+                requests: {
+                  cpu: "13.0",
```
zcain117, line 47: I tried just using:
```
cpu: "13.0",
memory: "40Gi",
```

but it did not reflect in thee final yaml

will-cromar: Hmmm, that shouldn't happen. Can you file a bug?

will-cromar (approved):

PR #299, opened 2020-09-30
Create a one shot running script and update the README.

allenwang28:
I originally meant to fix `gen-tests.sh` but it turns out that my macbook has a "open files limit" that my workstation doesn't so it wasn't broken.

Then I started writing this and now it exists.

### DIFF
--- a/scripts/run-oneshot.sh
+++ b/scripts/run-oneshot.sh
@@ -43,9 +43,6 @@ validate()
                              ;;
     "convergence" | "conv" ) testtype="conv"
                              ;;
-    * )                      echo "Invalid test type provided."
-                             exit
-                             ;;
   esac
 
   region=
@@ -105,7 +102,7 @@ run()
     jsonnet tests/oneshot.jsonnet -J . -S --tla-str test=$test_name | kubectl create -f -
   else
     echo "gcloud container clusters get-credentials oneshots-$region --region $region --project xl-ml-test"
-    echo "jsonnet tests/oneshot.jsonnet -J . -S --tla-str test=$test_name | kubectl create -f -"
+    echo "jsonnet tests/oneshot.jsonnet -J . -S --tla-str test=$test_name"
   fi
 }
 
--- a/scripts/run-oneshot.sh
+++ b/scripts/run-oneshot.sh
@@ -50,6 +50,8 @@ validate()
 
   region=
   case $accelerator in
+    # Pods are in europe-west4. Add in more cases here
+    # as necessary.
     "v2-32" | "v3-32" ) region="europe-west4"
                         ;;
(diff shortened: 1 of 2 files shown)


scripts/run-oneshot.sh
```diff
@@ -0,0 +1,135 @@
...
+    echo "Test type must be provided."
+    exit
+  fi
+
+  case $testtype in
+    "functional" | "func" )  testtype="func"
+                             ;;
+    "convergence" | "conv" ) testtype="conv"
+                             ;;
+    * )                      echo "Invalid test type provided."
```
will-cromar, line 46: This isn't quite accurate: some of our tests have custom modes like keras API and perfzero.

scripts/run-oneshot.sh
```diff
@@ -0,0 +1,135 @@
...
+  cd -
+
+  which jsonnet &> /dev/null
+  if [ -z "$dryrun" ]
+  then
+    gcloud container clusters get-credentials oneshots-$region --region $region --project xl-ml-test
+    jsonnet tests/oneshot.jsonnet -J . -S --tla-str test=$test_name | kubectl create -f -
+  else
+    echo "gcloud container clusters get-credentials oneshots-$region --region $region --project xl-ml-test"
+    echo "jsonnet tests/oneshot.jsonnet -J . -S --tla-str test=$test_name | kubectl create -f -"
```
will-cromar, line 108: Can dry run actually run jsonnet and show the YAML output instead of piping to kubectl?

scripts/run-oneshot.sh
```diff
@@ -0,0 +1,135 @@
...
+
+  set -e
+  set -x
+  cd -
+
+  which jsonnet &> /dev/null
+  if [ -z "$dryrun" ]
+  then
+    gcloud container clusters get-credentials oneshots-$region --region $region --project xl-ml-test
+    jsonnet tests/oneshot.jsonnet -J . -S --tla-str test=$test_name | kubectl create -f -
```
will-cromar, line 105: Do you think it would be useful to stream the logs back from this script automatically?

will-cromar (approved):
This is awesome! Thanks, Allen!

scripts/run-oneshot.sh
allenwang28: Yes, I think that would be a great addition!

scripts/run-oneshot.sh
allenwang28: Good point, I removed that path so any test type can be added.

scripts/run-oneshot.sh
allenwang28: Done

PR #300, opened 2020-10-01
Follow logs after starting oneshot job.

will-cromar:
Print jsonnet output for dry run.

Write YAML output to a temp file so user can inspect it later.


allenwang28 (approved):
Thanks!

zcain117 (approved):

PR #301, opened 2020-10-01
Email after n

zcain117:
Add a metrics handler config option "alert_after_second_test_failure". If true, then the metrics handler will wait for 2 failures in a row before sending an email alert for a failed test or a metrics regression.

Included unit tests cover almost all cases, but I also did some manual testing of the path that hits Bigquery. Tried deploying temporarily in prod and didn't see any issues


will-cromar (approved):
Thanks! This will be very helpful

PR #304, opened 2020-10-05
Remove XLNet Squad on v2-8 since it OOMs immediately

allenwang28:


tests/tensorflow/nightly/xlnet-squad.libsonnet
```diff
@@ -66,7 +65,7 @@ local tpus = import "templates/tpus.libsonnet";
   local v2_8 = {
```
will-cromar, line 65: Should you remove this mixin if there are no working v2-8 configs?

allenwang28: Oops, yes done. Also ran the gen script since I forgot.

k8s/us-central1/gen/tf-nightly-xlnet-squad-conv-v3-8.yaml
```diff
@@ -53,7 +53,6 @@
             - "--test_feature_path=$(XLNET_SQUAD_DIR)/spiece.model.slen-512.qlen-64.eval.features.pkl"
             - "--predict_file=$(XLNET_SQUAD_DIR)/dev-v2.0.json"
             - "--predict_dir=$(MODEL_DIR)"
-            - "--train_batch_size=48"
```
will-cromar, line 56: Is this change intentional?

will-cromar (approved):

k8s/us-central1/gen/tf-nightly-xlnet-squad-conv-v3-8.yaml
allenwang28: Yes, thanks for checking! I had --train_batch_size set in the model command but I prefer it with the accelerator mixin.

PR #306, opened 2020-10-05
Fix XLNet IMDB TFrecord path

allenwang28:
Didn't catch this before initial submission because it was failing due to a model level error. This just fixes `tfrecord` -> `tf_record`. Tested this via one shot.


will-cromar (approved):

zcain117 (approved):
Thanks!

PR #307, opened 2020-10-06
Updating dashboard to point TF-2.3 tests to TF-2.3.1 and adding tab for TF-2.2.1.

aman2930:


zcain117 (approved):
Just so you know - the dashboard gets data from our bigquery, which you can check in Cloud Console under our team's project. In bigquery, there are probably a bunch of rows that have `tf-r2.3` as the test name. With this PR, the effect would be to remove the tab of `tf-r2.3` tests and add a new tab to show `tf-r2.2.1` tests but this change won't migrate any of the `tf-r2.3` tests to the new tab. To do that, you'd have to update bigquery. I don't think you have to update bigquery unless this is what you wanted, I'm just pointing this out since the name of this PR implies that some tests are getting migrated to a new tab

aman2930:
Sorry for the confusing title. 
All 2.3 tests have already been using TF-2.3.1 now. So, now going forward there will be no new tests results for TF-2.3, but only for TF-2.3.1. 
Thanks for clearing that out.

PR #313, opened 2020-10-12
Add GPU ResourceQuota.

will-cromar:
Prevents GPU pods from entering the "Unschedulable" state when there's limited GPU quota. This should help reduce the number of timeouts.


zcain117:
Thanks for taking a look at this!

zcain117 (approved):

PR #315, opened 2020-10-13
Updating examples_per_second thresholds and timeouts

gagika:
Updating examples_per_second thresholds and timeouts for bert-mnli and classifier-resnet.


zcain117 (approved):
Worth a shot. Hopefully the stddev is not so tiny that this won't make a difference

will-cromar (approved):

PR #318, opened 2020-10-13
Model save load schedule

gagika:
I am updating the schedules Keras save/load tests so that the tests don't run the same time as they use the same gs buckets for testing.


tests/tensorflow/nightly/keras-api.libsonnet
```diff
@@ -147,7 +147,7 @@ local utils = import "templates/utils.libsonnet";
     keras_test + v3_8 + feature_column + timeouts.Hours(2),
     keras_test + v3_8 + preprocessing_layers,
     keras_test + v3_8 + rnn,
-    keras_test + v3_8 + save_and_load + timeouts.Hours(2),
+    keras_test + v3_8 + save_and_load + timeouts.Hours(2) + {schedule: "30 12 * * *"},
```
zcain117, line 150: woah cool syntax

zcain117 (approved):

PR #330, opened 2020-10-27
Pin r1.7 HF and fix nightly HF GLUE task

jysohn23:
On PT/XLA examples repo I've also updated the pin to HF dep to their `v3.4.0` release.

### DIFF
--- a/tests/pytorch/nightly/hf-glue.libsonnet
+++ b/tests/pytorch/nightly/hf-glue.libsonnet
@@ -22,12 +22,12 @@ local utils = import "templates/utils.libsonnet";
     git clone https://github.com/huggingface/transformers.git
     cd transformers && pip install .
     git log -1
+    pip install datasets
     python examples/xla_spawn.py \
       --num_cores 8 \
       examples/text-classification/run_glue.py \
       --logging_dir=./tensorboard-metrics \
       --task_name MNLI \
-      --data_dir /datasets/glue/MNLI \
       --cache_dir ./cache_dir \
       --do_train \
       --do_eval \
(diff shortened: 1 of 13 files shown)


tests/pytorch/nightly/hf-glue.libsonnet
```diff
@@ -22,12 +22,12 @@ local utils = import "templates/utils.libsonnet";
...
+    pip install datasets
     python examples/xla_spawn.py \
       --num_cores 8 \
       examples/text-classification/run_glue.py \
       --logging_dir=./tensorboard-metrics \
       --task_name MNLI \
-      --data_dir /datasets/glue/MNLI \
       --cache_dir ./cache_dir \
       --do_train \
       --do_eval \
```
zcain117, line 33: could you also remove from down below:

```
    volumeMap+: {
      datasets: common.datasetsVolume,
    },
```

tests/pytorch/r1.7/hf-glue.libsonnet
```diff
@@ -19,8 +19,8 @@ local utils = import "templates/utils.libsonnet";
 
 {
   local command_common = |||
-    git clone https://github.com/huggingface/transformers.git
-    cd transformers && pip install .
+    git clone --recursive -b r1.7 https://github.com/pytorch-tpu/examples.git
+    cd examples/deps/transformers && pip install .
     git log -1
     python examples/xla_spawn.py \
       --num_cores 8 \
```
zcain117, line 26: could you also remove from down below:

```
    volumeMap+: {
      datasets: common.datasetsVolume,
    },
```

zcain117 (changes requested):
Thanks for this fix!

tests/pytorch/nightly/hf-glue.libsonnet
jysohn23: Good point, done.

tests/pytorch/r1.7/hf-glue.libsonnet
jysohn23: `r1.7` depends on `v3.4.0` (latest stable) which still doesn't have the latest on demand GLUE dataset download change so we need the volume.

zcain117 (approved):

PR #334, opened 2020-11-03
Add EfficientNet-b7 functional test

allenwang28:


allenwang28:
https://pantheon.corp.google.com/kubernetes/job/us-central1/oneshots-us-central1/default/tf-nightly-classifier-efficientnet-hbm-v3-8-2ldkg/details?project=xl-ml-test

tests/tensorflow/nightly/classifier-efficientnet.libsonnet
```diff
@@ -46,11 +51,32 @@ local gpus = import "templates/gpus.libsonnet";
...
   },
+  local functional_b7 = common.Functional {
+    mode: "hbm",
+    paramsOverride+: {
+      train+: {
+        epochs: 1, 
+      },
+      model+: {
+        model_params+: {
+          model_name: "efficientnet-b7",
```
will-cromar, line 62: Would it be better to make separate mixins for efficientnets b0 and b7? E.g.

```
local b0 = {
    paramsOverride+: {
      model+: {
        model_params+: {
          model_name: "efficientnet-b0",
        },
      },
  }
}
```

Usage would be like `efficientnet + b0 + v3_8 + functional`

allenwang28: Good idea - I just changed it so that the model name defaults to efficientnet-b0 but is overridable.

I think your approach is best but requires some additional tinkering for the `mode` field. I don't currently have plans to add other model tests so I just renamed the b7 tests to "hbm" test to be clearer.

will-cromar: You can also add something like this to the mixin: `local b0 = { modelName+: "-b0", ... }`. But renaming the mode to `hbm` also works

allenwang28: Oops - didn't actually push the comment related to the last comment. Should be updated now!

will-cromar (approved):

PR #335, opened 2020-11-03
Rescheduling tests due to limited TPU resources. 

aman2930:
- Turned down tests for TF-2.1.2, TF-2.2.1. 
- Functional tests for TF-1.15.4 
- Functional test for TF-1.15.4 on Monday and Wednesday @ 10PM PST (Twice a week).
- Convergence test for TF-1.15.4 on Sunday at Midnight (Once a week).
- Functional test for TF-2.3.1 on Tuesday and Thursday @ 10PM PST (Twice a week).
- Convergence test for TF-2.3.1 on Saturday at Midnight (Once a week).


tests/tensorflow/r1.15/common.libsonnet
```diff
@@ -27,8 +27,8 @@ local mixins = import "templates/mixins.libsonnet";
   Convergence:: mixins.Convergence {
     schedule: "0 8 * * 0"
   },
-  # Running convergence tests at 10PM PST daily.
+  # Running functional tests at 10PM PST on Monday, Wednesday..
   Functional:: mixins.Functional {
-    schedule: "0 6 * * *"
```
will-cromar, line 32: Can we disable these functional tests by default? We're already running the convergence tests, and we don't need the quick feedback while we're not testing RCs.

We can override one or two short tests to run nightly if we want a frequent sanity check. E.g. you can override a schedule like this: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/tensorflow/r1.15/resnet.libsonnet#L56

tests/tensorflow/r2.3/common.libsonnet
```diff
@@ -40,14 +40,15 @@ local mixins = import "templates/mixins.libsonnet";
       },
     },
   },
-  # Don't run tests manually by default since this release is stable.
+  # Running convergence tests at 10 PM PST on Tuesday, Thursday.
   Functional:: mixins.Functional {
-    schedule: null,
+    schedule: "0 6 * * 2,4",
```
will-cromar, line 45: Same for 2.3.

will-cromar (approved):

PR #340, opened 2020-11-11
Skip rows with nan as value. Use math.isnan to catch some versions of…

zcain117:
… NaN

TIL:
```
>>> a = [math.nan]
>>> math.nan in a
True
>>> float('nan') in a
False
>>> math.isnan(math.nan)
True
>>> math.isnan(float('nan'))
True
```

So we should use `math.isnan` to catch NaN that has been parsed using `float`

Also skip rows entirely if the metric value is NaN since they will just be noise

### DIFF
--- a/metrics_handler/util_test.py
+++ b/metrics_handler/util_test.py
@@ -65,8 +65,11 @@ class UtilTest(parameterized.TestCase):
 
   @parameterized.named_parameters(
     ('inf', [1, 2, 3, math.inf, 4, 5], [1, 2, 3, None, 4, 5]),
+    ('inf_parsed', [1, 2, 3, float('inf'), 4, 5], [1, 2, 3, None, 4, 5]),
     ('-inf', [1, 2, 3, -math.inf, 4, 5], [1, 2, 3, None, 4, 5]),
     ('nan', [1, 2, 3, math.nan, 4, 5], [1, 2, 3, None, 4, 5]),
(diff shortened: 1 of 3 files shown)


metrics_handler/util_test.py
```diff
@@ -65,8 +65,11 @@ def test_workload_link(self):
 
   @parameterized.named_parameters(
     ('inf', [1, 2, 3, math.inf, 4, 5], [1, 2, 3, None, 4, 5]),
+    ('inf_parsed', [1, 2, 3, float('inf'), 4, 5], [1, 2, 3, None, 4, 5]),
     ('-inf', [1, 2, 3, -math.inf, 4, 5], [1, 2, 3, None, 4, 5]),
     ('nan', [1, 2, 3, math.nan, 4, 5], [1, 2, 3, None, 4, 5]),
+    ('nan_parsed', [1, 2, 3, float("nan"), 4, 5], [1, 2, 3, None, 4, 5]),
+    ('nan_cap_parsed', [1, 2, 3, float("NAN"), 4, 5], [1, 2, 3, None, 4, 5]),
```
will-cromar, line 72: I assume this check is necessary but 😬

will-cromar (approved):
Wow. TIL. This is not something I expected to be a problem. Thanks for the quick fix

will-cromar:
I didn't want to believe it, so I tried it myself:

```
>>> float('NaN') == float('nan')
False
>>> float('inf') == float('Inf')
True
```

PR #349, opened 2020-12-10
Reservation tests

zcain117:
Split into 2 initial commits to make this more readable:
1. updates to templates and test files, 2 example yaml changes - 1 using reservation and 1 not using reservation
2. rest of the yaml updates

I chose to run TF nightly and TF 1.15 v2-32 convergence tests on our v2-32 reservation, trying to keep utilization high. I set all the `nightly` tests to run at an hour of day >= 7, since our nightly image right now gets published at ~5 in UTC time. The 1.15 tests can start at any time of day since that image is pretty static.

I generated the cron schedules with a helper python script (not checked in) but looked over them manually too. I'll save the script somewhere to make it easy for us to set up a new schedule if we want to change the tests that run reserved.

For testing: I upgraded version on oneshot cluster and ran a test there using reservation + verified that the TPU was actually from our reservation using `gcloud compute tpus describe` while the job was running

If this PR looks OK, I will:
1. upgrade real pods cluster
2. push one of these reservation yamls
3. run test job on real cluster
4. submit the PR if the real job looks OK

### DIFF
--- a/templates/tpus.libsonnet
+++ b/templates/tpus.libsonnet
@@ -28,13 +28,14 @@ local base = import 'base.libsonnet';
       metadata: {
         annotations: {
           "tf-version.cloud-tpus.google.com": tpuSettings.softwareVersion,
+	  "reserved.cloud-tpus.google.com": tpuSettings.reserved,
         },
       },
       spec+: {
         containerMap+: {
           train+: {
             resources+: {
-              local preemptiblePrefix = 
+              local preemptiblePrefix =
                 if tpuSettings.preemptible then
                   "preemptible-"
                 else
--- a/templates/tpus.libsonnet
+++ b/templates/tpus.libsonnet
@@ -66,5 +66,4 @@ local base = import 'base.libsonnet';
   v3_8: self.TpuSpec { version: 3, size: 8 },
   v2_32: self.TpuSpec { version: 2, size: 32 },
   v3_32: self.TpuSpec { version: 3, size: 32 },
-
 }
--- a/templates/tpus.libsonnet
+++ b/templates/tpus.libsonnet
@@ -25,10 +25,11 @@ local base = import 'base.libsonnet';
     replicas: tpu.size / 8, # Each TPU replica has 8 cores
 
     PodTemplate(tpuSettings):: {
+      assert !(tpuSettings.preemptible == "true" && tpuSettings.reserved == "true"): ("TPU cannot be both reserved and preemptible"),
       metadata: {
         annotations: {
           "tf-version.cloud-tpus.google.com": tpuSettings.softwareVersion,
-	  "reserved.cloud-tpus.google.com": tpuSettings.reserved,
+          "reserved.cloud-tpus.google.com": tpuSettings.reserved,
         },
       },
       spec+: {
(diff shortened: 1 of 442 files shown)


tests/tensorflow/nightly/classifier-efficientnet.libsonnet
```diff
@@ -158,7 +163,7 @@ local gpus = import "templates/gpus.libsonnet";
     efficientnet + v3_8 + convergence + timeouts.Hours(45),
     efficientnet + v2_32 + functional,
     efficientnet + v3_32 + functional,
-    efficientnet + v2_32 + convergence + timeouts.Hours(30),
+    efficientnet + v2_32 + convergence + timeouts.Hours(10) + reserved + {schedule: "0 7 * * 1,3,5,6"},
```
will-cromar, line 166: Did you mean to change this timeout?

tests/tensorflow/nightly/classifier-efficientnet.libsonnet
```diff
@@ -144,6 +144,11 @@ local gpus = import "templates/gpus.libsonnet";
   local v3_32 = tpu_common {
     accelerator: tpus.v3_32,
   },
+  local reserved = {
```
will-cromar, line 147: This might be better as a mixin in the `tpus.libsonnet` file. The usage would be `... + tpus.reserved. We used to have something similar for preemptible

templates/tpus.libsonnet
```diff
@@ -28,13 +28,14 @@ local base = import 'base.libsonnet';
       metadata: {
         annotations: {
           "tf-version.cloud-tpus.google.com": tpuSettings.softwareVersion,
+	  "reserved.cloud-tpus.google.com": tpuSettings.reserved,
```
will-cromar, line 31: I'm surprised this is different than the implementation for preemtible (ie with a resource name like `reserved-v3`). That would have made it easy to implement a limit with a ResourceQuota

will-cromar: This also means that you could actually request a preemptible reserved TPU pod, which doesn't make any sense. Can you add a check to raise an error in that case?

will-cromar (approved):
Overall LGTM. Just minor nits.

The schedules are fine for now, but long-term we should implement a ResourceQuota to limit the number of reserved cores in use to be the same as our project quota. We can do that with PriorityClasses like we do for GPUs: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/68c67c32b203fe332c6d2068f11a83742801120d/k8s/us-central1/quota.yaml#L49

tests/tensorflow/nightly/classifier-efficientnet.libsonnet
zcain117: yes - this test has taken ~8hrs for the last few months. That last time it was over 10 hours was in July

tests/tensorflow/nightly/classifier-efficientnet.libsonnet
zcain117: good idea - will fix

templates/tpus.libsonnet
zcain117: sure, I'll add an `assert`

zcain117:
upgraded pods cluster and ran one of the reserved tpu jobs on it. Verified the TPU pods was pulled from reservation

PR #351, opened 2020-12-15
Add new metrics format to templates

will-cromar:
Add temporary hack to convert from old format to new format.

@zcain117 FYI


will-cromar:
There's a huge number of generated lines in this commit. The major change is here: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/pull/351/commits/c35343f4a45f0a16b145cfb9eef621a7767b503b

zcain117 (approved):

PR #368, opened 2021-01-20
Add actions to diff and build generated files.

will-cromar:
This change will remove the need for contributors to rebuild generated files. Each PR will have a check showing the diff that contributors can use as a sanity-check. Generated files will be built for each push to master and kept on the gen branch. After this PR is submitted, I can update Cloud Build to use the generated files from gen instead.

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -1,0 +1,40 @@
+# Build generated files with JSonnet and show diff output.
+name: Diff Generated Files
+
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


zcain117:
The check failed:
```
Run git add -N k8s/
fatal: couldn't find remote ref gen
Error: Process completed with exit code 128.
```

will-cromar:
Fixed. I had to create the gen branch in this repository.

zcain117 (approved):

PR #371, opened 2021-01-21
Moving TF-2.4.0 tests to TF-2.4.1 for Patch release regression.

aman2930:


will-cromar:
Do you want to add the dashboard tab in this PR as well?

aman2930:
Added the dashboard.

will-cromar (approved):
LGTM

jysohn23:
Seems like this didn't run `./scripts/gen-tests.sh`? At master head, if you run that script it's picking up all those changes.

jysohn23:
> Seems like this didn't run `./scripts/gen-tests.sh`? At master head, if you run that script it's picking up all those changes.

Just chatted w/ @zcain117 looks like we now have postsubmit run the above and yamls are no longer accurate on master HEAD.

PR #378, opened 2021-02-18
switching to xla_dist for resnet50 functional test

sshahrokhi:


tests/pytorch/nightly/resnet50-pod.libsonnet
```diff
@@ -25,6 +25,13 @@ local mixins = import "templates/mixins.libsonnet";
       "/pytorch/xla/test/test_train_mp_imagenet.py",
     ],
   },
+  local resnet50_pod_func = common.PyTorchXlaDistPodTest {
+    modelName: "resnet50-pod",
+    command: [
+      "python3",
+      "/usr/share/torch-xla-nightly/pytorch/xla/test/test_train_mp_imagenet.py",
```
zcain117, line 32: does this path exist on the test machine?  It will be running inside a docker container

sshahrokhi: I ran it with oneshot and it worked, I think the correct path is this if we use xla_dist.

zcain117 (approved):

PR #386, opened 2021-03-03
Remove $ from cluster_name in stackdriver query

zcain117:


jysohn23 (approved):
Thanks @zcain117!

will-cromar (approved):
Thanks!

PR #388, opened 2021-03-04
remove cifar tests

zcain117:
I'll delete the workloads manually in kubernetes

@will-cromar anything else to do to prevent them from being regenerated?


jysohn23 (approved):
Thanks!

will-cromar (approved):
Deleting the workloads is all you need to do

PR #395, opened 2021-03-25
Add wav2vec2 TPU tests

taylanbil:


tests/pytorch/nightly/wav2vec2.libsonnet
```diff
@@ -0,0 +1,130 @@
...
+      datasets: common.datasetsVolume,
+    },
+    jobSpec+:: {
+      template+: {
+        spec+: {
+          containerMap+: {
+            train+: {
+              resources+: {
+                requests: {
+                  cpu: "40.0",
```
zcain117, line 97: could you try making these top-level like here: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/pytorch/nightly/fs-transformer.libsonnet#L61

and see if it changes the `.yaml` at all?  I think it should translate to the same yaml

taylanbil:
Functional test passed, fired up a conv. test.

cc @bilgeacun @alexeib

zcain117 (approved):
LGTM assuming the oneshots pass

taylanbil:
Convergence oneshot test passed.

PR #399, opened 2021-04-02
Adding tests for TF-2.5.0.

aman2930:
Reducing the frequency of TF-2.4 tests so that we do have enough resources to carry out TF-2.5 tests.


will-cromar (approved):
LGTM

PR #405, opened 2021-04-20
Extend timeout for pt-nightly python ops test.

zcain117:


zcain117:
I tried this manually using the TPUVM test (it was using pytorch xla nightly)  and it completed the whole test suite, so I think this new timeout is long enough

JackCaoG (approved):
Wow.. op test now takes a long time to run. This reminds me that I observe that op test is slower on tpuvm comparing to 2vm. I am wondering if running the server in a separate process will help (don't need to recreating server over and over).

will-cromar (approved):

zcain117:
> Wow.. op test now takes a long time to run. This reminds me that I observe that op test is slower on tpuvm comparing to 2vm. I am wondering if running the server in a separate process will help (don't need to recreating server over and over).

I haven't yet confirmed any speed difference between TPUVM and 2VM. I was assuming they are similar but I'll let you know

PR #428, opened 2021-06-03
refactoring jax tests and adding latest and nightly versions

sshahrokhi:
refactoring the flax tests and adding two versions (latest and nightly)


will-cromar (approved):
LGTM

PR #433, opened 2021-06-04
Add TPUVM pod test and add xla_dist test.

zcain117:
Ran the TPUVM v3-32 test for both resnet and mnist and both succeeded.

For the 2VM `xla_dist` test: no way to test yet since the conda env hasn't rolled out yet. I'll monitor that test and hopefully it will start working once the conda env rolls out tomorrow


JackCaoG (approved):
Thanks @zcain117 !

PR #445, opened 2021-06-16
Enabling functional tests for TF-2.2.3 regression.

aman2930:


tests/tensorflow/r2.2/mnist.libsonnet
```diff
@@ -26,7 +26,7 @@ local tpus = import 'templates/tpus.libsonnet';
       '--model_dir=$(MODEL_DIR)',
     ],
   },
-  local convergence = common.Convergence {
+  local functional = common.Functional {
```
will-cromar, line 29: This is causing a build error

aman2930: Fixed.

will-cromar (approved):
Thanks for adding these tests

PR #446, opened 2021-06-17
Add r1.9 python TPUVM op tests. Still using manual install becaause T…

zcain117:
…PUVMs still come with 1.8.1

Pushed workloads to k8s and confirmed they start up successfully and begin testing. If any tests fail we'll modify the underlying test, not this yaml

### DIFF
--- a/tests/pytorch/r1.9/python-ops.libsonnet
+++ b/tests/pytorch/r1.9/python-ops.libsonnet
@@ -13,8 +13,10 @@
 // limitations under the License.
 
 local common = import 'common.libsonnet';
+local experimental = import '../experimental.libsonnet';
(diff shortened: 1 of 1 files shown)


will-cromar (approved):

tests/pytorch/r1.9/python-ops.libsonnet
```diff
@@ -13,8 +13,10 @@
 // limitations under the License.
 
 local common = import 'common.libsonnet';
```
will-cromar, line 15: Please sort these imports

will-cromar: Resolved

zcain117: done

PR #456, opened 2021-06-25
Delete Kubernetes workloads automatically when tests are removed.

will-cromar:


zcain117 (approved):
Great idea, thanks Will !

PR #460, opened 2021-06-29
Fix jaxlib install command.

skye:
This should fix the jaxlib-head tests.


sshahrokhi (approved):

skye:
Can we merge this?

skye:
I should clarify, I don't have permission to do so.

PR #466, opened 2021-07-12
Update jax[tpu] install command

skye:


skye:
@sshahrokhi

skye:
Actually please don't merge yet, I think this still isn't right... I failed to fully document how the jax[tpu] install is supposed to work and something is still missing.

PR #474, opened 2021-07-19
Adjust threshold for eval loss to match historic value.

zcain117:
Loss has been stable around 2.8 for this test since q4 2020


JackCaoG (approved):
Thanks @zcain117 !

PR #476, opened 2021-07-21
Use the TF version that ships with the VM (2.6.0) on nightly jobs.

skye:
Tests currently pip install TF 2.5.0 to avoid a bunch of segfaults and other errors when using JAX + TF. This seems to be resolved in the nightly image, but using TF 2.5.0 causes errors like:

```
/usr/local/lib/python3.8/dist-packages/tensorflow/core/kernels/libtfkernel_sobol_op.so: undefined symbol: _ZNK10tensorflow8OpKernel11TraceStringB5cxx11ERKNS_15OpKernelContextEb"
```

### DIFF
--- a/tests/jax/common.libsonnet
+++ b/tests/jax/common.libsonnet
@@ -37,6 +37,7 @@ local tpus = import 'templates/tpus.libsonnet';
       maybeBuildJaxlib: error 'Must define `maybeBuildJaxlib`',
       installLocalJax: error 'Must define `installLocalJax`',
       installLatestJax: error 'Must define `installLatestJax`',
+      maybeInstallTF: error 'Must define `maybeInstallTF`',
       printDiagnostics: |||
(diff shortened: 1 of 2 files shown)


sshahrokhi:
Would this fix the nightly problem you mentioned? then maybe we need to remove this line as well: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/797df3a785e919c247bda8e03d0bdc7fa155c46e/tests/jax/nightly/common.libsonnet#L45

skye:
Oh you're right, I think so! I'm trying it now, but we could also just merge and see.

skye:
It still fails actually. This probably warrants more investigation, but I'm just keeping the nightly tests as-is for now.

sshahrokhi (approved):

sshahrokhi:
bummer that it didn't fix that!

PR #481, opened 2021-07-26
Adjust the bound for resnet50 aten calls TPUVM test.

zcain117:
Double checked what the true expected value should be


JackCaoG (approved):
Thanks!

PR #486, opened 2021-07-29
Add test for nightly TPUVM for fairseq transformer.

zcain117:
I ran this once and it took over a day to schedule. It failed but the reason it failed should be fixed by https://github.com/GoogleCloudPlatform/ml-testing-accelerators/pull/485

I'd like to submit it and will debug if it fails during its normal automated run


JackCaoG (approved):
Thanks @zcain117 !

PR #487, opened 2021-07-30
Update pytorch 1.8.1 tests to install 1.8.1 wheels during setup.

zcain117:
https://b.corp.google.com/issues/195062592

### DIFF
--- a/tests/pytorch/r1.8.1/common.libsonnet
+++ b/tests/pytorch/r1.8.1/common.libsonnet
@@ -52,4 +52,14 @@ local volumes = import 'templates/volumes.libsonnet';
     name: 'pytorch-datasets-claim',
     mountPath: '/datasets',
   },
+  tpu_vm_1_8_1_install: |||
+    sudo pip3 uninstall --yes torch torch_xla torchvision
+    sudo pip3 install torch==1.8.1
+    sudo pip3 install torchvision==0.9.1
+    sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/torch_xla-1.8.1-cp36-cp36m-linux_x86_64.whl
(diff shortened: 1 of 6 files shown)


tests/pytorch/r1.8.1/common.libsonnet
```diff
@@ -52,4 +52,14 @@ local volumes = import 'templates/volumes.libsonnet';
...
   },
+  tpu_vm_1_8_1_install: |||
+    sudo pip3 uninstall --yes torch torch_xla torchvision
+    sudo pip3 install torch==1.8.1
+    sudo pip3 install torchvision==0.9.1
+    sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/torch_xla-1.8.1-cp36-cp36m-linux_x86_64.whl
+    git clone https://github.com/pytorch/pytorch.git -b release/1.8.1
+    cd pytorch
+    git clone https://github.com/pytorch/xla.git -b r1.8.1
+    export XRT_TPU_CONFIG='localservice;0;localhost:51011'
```
jysohn23, line 63: Is there anywhere we put setup stuff that's shared for all TPU VM PyTorch/XLA? Maybe XRT_TPU_CONFIG would fit better there, but no worries if not.

jysohn23 (approved):

tests/pytorch/r1.8.1/common.libsonnet
zcain117: yeah we already do export it in the global Pytorch TPUVM setup here: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/pytorch/experimental.libsonnet#L46

So I will remove it from the 1.8.1 specific setup

PR #492, opened 2021-08-05
Use v2-alpha for the stable versions of PyTorch for TPUVM tests.

zcain117:
all our TPUVM tests were using v2-nightly, but it makes more sense for the stable versions to run on v2-alpha. Nightly wheels will keep using v2-nightly for TPUVM version


JackCaoG (commented):
Thank for doing this! It is large change, did you have a chance to run an experiment to verify it will work?

will-cromar:
Please re-open if this is still an issue.

PR #493, opened 2021-08-18
Updated the fairseq-cli to use hydra-config file

ultrons:
fairseq-cli in it's current form does not support mask related options.
However these options are supported via config file.
Updated the test to use config file.


ultrons:
> 

@will-cromar is the original test passing now?
My understanding was that these updates would still be required or we are not running this test? 
Please confirm.

will-cromar:
Reopening because this test is still broken

miaoshasha:
testing this change

JackCaoG:
@will-cromar Can we merge this change? I think I fixed the w2v2 so with this change it should start running. If it failed again I will take another pass.

will-cromar (approved):

PR #504, opened 2021-09-13
Prune TensorFlow tests.

will-cromar:
Per our discussion last week, removing many TensorFlow tests that are not useful anymore. Also removing Keras API tests for past patches.

PR #505, opened 2021-09-14
Remove old metrics format from templates

will-cromar:


will-cromar:
This is effectively a no-op for PyTorch tests and very close to a no-op for the TF tests. The main change to the generated files is removing the old publisher container, which is currently ignored already.

Waiting to submit until after #504 goes through

sshahrokhi (approved):
Thanks for all the changes! I just was wondering why you didn't put the examples_per_second assertion in the main common before the versions (and for Functional and Convergence tests) but I saw that the standard deviation was different for a few of them, so I thought that is fine!

will-cromar:
Thanks for reviewing this huge PR.

Yeah, the organization of our metric assertions could use some refactoring, but I chose to stay as close to the original code as I could with the new format.

PR #512, opened 2021-09-22
use var in common

miaoshasha:


miaoshasha:
@googlebot I signed it!

PR #514, opened 2021-09-23
Fix TPU creation retry loop

skye:
Without this, the script will exit on the first creation failure.

### DIFF
--- a/tests/experimental.libsonnet
+++ b/tests/experimental.libsonnet
@@ -98,6 +98,7 @@ local volumes = import 'templates/volumes.libsonnet';
 
               # Retry every 30 seconds for 10 minutes
               for i in {1..20}; do
+                set +e
                 gcloud alpha compute tpus tpu-vm create ${tpu_name} \
                   --accelerator-type=%(acceleratorName)s \
                   --version=%(softwareVersion)s  \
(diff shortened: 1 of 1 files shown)


tests/experimental.libsonnet
```diff
@@ -107,6 +108,7 @@ local volumes = import 'templates/volumes.libsonnet';
 
                 exit_code=$?
                 test $exit_code = 0 && break || sleep 30;
+                set -e
```
will-cromar, line 111: Should this go up one line? Otherwise when the loop breaks, we'll still have `+e` set

will-cromar (commented):
🤦‍♂️ Thanks for fixing this

tests/experimental.libsonnet
skye: Great call. I just put it around the gcloud command.

will-cromar: It should go after `exit_code=$?`, otherwise exit_code will always be 0 because `set -e` will alway work:

```
wcromar-macbookpro2:~ wcromar$ set +e
wcromar-macbookpro2:~ wcromar$ fail
-bash: fail: command not found
wcromar-macbookpro2:~ wcromar$ set -e
wcromar-macbookpro2:~ wcromar$ echo $?
0
wcromar-macbookpro2:~ wcromar$ 
```

skye: omg. I swear I thought of this the first time around :P Done.

will-cromar (approved):

PR #516, opened 2021-09-27
Remove duration assertion from JAX tests.

will-cromar:
Per our discussion offline

Also remove some obsolete config I missed in #505


sshahrokhi (approved):

skye (approved):
Thanks!

PR #518, opened 2021-10-01
R1.9.1

miaoshasha:
create tests for 1.9.1 version, mock, for checking if tests can be correctly created. will delete later
the 1.9.1 wheels are incorrectly named as 1.9, but notice the build date 09/23. so they actually are 1.9.1 wheels (copies at build time).

### DIFF
--- a/tests/pytorch/experimental.libsonnet
+++ b/tests/pytorch/experimental.libsonnet
@@ -191,8 +191,6 @@ local experimental = import '../experimental.libsonnet';
       tpuVmStartupScript: |||
         sudo bash /var/scripts/docker-login.sh
         sudo docker rm libtpu || true
-        sudo docker create --name libtpu gcr.io/cloud-tpu-v2-images/libtpu:pytorch-1.9.1 "/bin/bash"
-        sudo docker cp libtpu:libtpu.so /lib
         sudo pip3 uninstall --yes torch torch_xla torchvision numpy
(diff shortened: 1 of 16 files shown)


miaoshasha:
testing locally with:
scripts/run-oneshot.sh --accelerator v2-8 --file tests/tensorflow/nightly/mnist.libsonnet --type functional

tests/pytorch/experimental.libsonnet
```diff
@@ -184,4 +184,73 @@ local experimental = import '../experimental.libsonnet';
...
     },
   },
+  PyTorch1_9_1TpuVmPodTest:: experimental.BaseTpuVmMixin {
+    local config = self,
+    tpuSettings+: {
+      softwareVersion: 'v2-nightly',
+      tpuVmStartupScript: |||
+        sudo bash /var/scripts/docker-login.sh
+        sudo docker rm libtpu || true
+        sudo docker create --name libtpu gcr.io/cloud-tpu-v2-images/libtpu:pytorch-1.9.1 "/bin/bash"
```
jysohn23, line 194: This image doesn't seem to exist?

miaoshasha:
local run of oneshot script succeeds, job suffix 5k7bn. @jysohn23

jysohn23:
Could you check with Will where the `tpuVmStartupScript` logs are? We should verify that `libtpu` from pypi is being installed properly. This is probably how you want to install with extra dependencies on libtpu by asking to optionally install with `tpuvm` support.

PR #532, opened 2021-11-10
Install tensorflow-text and keras into the XLML test image.

aman2930:


images/tensorflow/Dockerfile
```diff
@@ -21,9 +21,13 @@ RUN apt-get update && apt-get install -y --no-install-recommends git
 
 RUN pip3 install cloud-tpu-client pyyaml
 
+ARG tensorflow_text_version=latest
+ARG keras_version=latest
+
 # HACKs: Remove this once they are added to the Model Garden requirements
 RUN pip3 install tensorflow-recommenders --no-deps
-RUN pip3 install tensorflow-text==2.6.0rc0 --no-deps
+RUN pip3 install tensorflow-text==${tensorflow_text_version} --no-deps
```
will-cromar, line 29: The `tensorflow-text==latest` syntax doesn't work:

```
root@822832b71ff2:/# pip install tensorflow-text==latest
ERROR: Could not find a version that satisfies the requirement tensorflow-text==latest
ERROR: No matching distribution found for tensorflow-text==latest
```

images/tensorflow/Dockerfile
```diff
@@ -21,9 +21,13 @@ RUN apt-get update && apt-get install -y --no-install-recommends git
...
+RUN pip3 install keras==${keras_version} --no-deps
```
will-cromar, line 30: Is this necessary? And will it work for nightly? New nightly images have keras-nightly installed already

will-cromar (changes requested):

images/tensorflow/Dockerfile
aman2930: Fixed it.

images/tensorflow/Dockerfile
aman2930: Removed it. 
Actual error is that while installing requirements.txt packages, it is indirectly installing TF-2.6.0. Now it has been fixed.

images/tensorflow/Dockerfile
```diff
@@ -21,15 +21,22 @@ RUN apt-get update && apt-get install -y --no-install-recommends git
...
+    pip3 install tensorflow-text==${tensorflow_text_version} --no-deps; \
+    fi
 
+ 
 # Checkout tensorflow/models at HEAD
 WORKDIR /
 RUN curl -L https://github.com/tensorflow/models/archive/${model_garden_branch}.tar.gz | tar zx
 RUN mv models-${model_garden_branch} garden/
-RUN pip3 install -r /garden/official/requirements.txt
+RUN pip3 install -r /garden/official/requirements.txt --no-deps
```
will-cromar, line 39: Have you tested this build with the `--no-deps` flag?

images/tensorflow/cloudbuild.yaml
```diff
@@ -30,5 +31,6 @@ substitutions:
   _VERSION: 'latest'
   _BASE_IMAGE_VERSION: 'nightly'
   _MODEL_GARDEN_BRANCH: 'master'
+  _TENSORFLOW_TEXT_VERSION: 'latest'
```
will-cromar, line 34: This should probably be 'nightly'

images/tensorflow/Dockerfile
aman2930: Yes. That is something I was testing in the dashboard. It seems like --no-deps is not working out for us. 
Next option is to fix requirements.txt with proper version of the package which is indirectly installing 2.6.0.

aman2930: I have fixed the requirements.txt file in the model garden repo with commit: https://github.com/tensorflow/models/commit/31fd5b760458add99709c0b646381224355c008f#diff-6896191a14f6c66534bac457f50996b9330cd702cb6dbaae4c08d1d213e93d98

will-cromar (approved):

PR #542, opened 2021-12-16
fix resnet pod test to use pod vm and xla_dist

miaoshasha:
use 32 cores instead of 8 cores which is the current setup
tested generated yaml file, model trains well


will-cromar (approved):

will-cromar:
Please run `scripts/format-jsonnet.sh`

PR #550, opened 2022-01-11
Refactor `fs-transformer` to enable v4 version of the test

will-cromar:


will-cromar:
I made the `PyTorchTpuVmMixin` more like the JAX one, where we can substitute in extra setup steps as needed without having to change the model run command. I also moved the epic stack of runtime flags into a list to clean up the messy string manipulation. This change will make it easier to update settings for the v4 version of the test.

Checked the diff. Most changes are either whitespace or rearranging flags. This should be a no-op for existing tests.

templates/tpus.libsonnet
```diff
@@ -22,6 +22,7 @@ local base = import 'base.libsonnet';
     type: 'tpu',
     version: error 'Must override `version`',
     size: error 'Must override `size`',
+    numCores: if tpu.version <= 3 then 8 else 4,
     replicas: tpu.size / 8,  // Each TPU replica has 8 cores
```
allenwang28, line 26: do we also need to change this?

`replicas: tpu.size / self.numCores`

allenwang28 (changes requested):
Thanks for cleaning this up! This should be a great reference for others

templates/tpus.libsonnet
will-cromar: This calculation should still be correct -- `replicas` essentially means hosts in this context.

allenwang28 (approved):

PR #563, opened 2022-02-09
Clean up deprecated code.

will-cromar:
Remove `ci_pytorch` and `docs/tutorial` since both of these are outdated and broken. Bugs have already been filed to update them.


will-cromar:
There is no impact to the generated test files.

PR #574, opened 2022-02-18
add pytorch 1.11 tests

miaoshasha:
pytorch:
add 1.11 tests, remove 1.9 tests

To be merged after RCs are created.


tests/pytorch/r1.11/python-ops.libsonnet
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 Google LLC
+// Copyright 2020 Google LLC
```
JackCaoG, line 1: I am guessing we should change it to 2022?

tests/pytorch/r1.9/colab.libsonnet
```diff
@@ -1,55 +0,0 @@
-local common = import 'common.libsonnet';
```
will-cromar, line 1: Can you include the colab notebook tests in r1.11?

tests/pytorch/r1.11/python-ops.libsonnet
miaoshasha: done

tests/pytorch/r1.9/colab.libsonnet
miaoshasha: added

tests/pytorch/r1.11/python-ops.libsonnet
JackCaoG: All of the files in this pr seems to share this typo, can you fix them?

miaoshasha: done

JackCaoG (approved):
I will approve this pr to kick of test for now. If there is anything we need to change we can do that in the following pr.

PR #576, opened 2022-02-25
Use `nightly_3.7` docker tag for all nightly PyTorch TPU node tests

will-cromar:
Previously, they were using the `nightly` tag, which stopped updating in December 2021.

Consolidate MNIST tests.

Fix entrypoint for PyTorch so we don't have to `unset XRT_TPU_CONFIG` in GPU tests.


JackCaoG:
Hey Will, I am not as familiar this part of the config, do you mind explaining a bit more?

tests/pytorch/nightly/common.libsonnet
```diff
@@ -20,27 +20,16 @@ local volumes = import 'templates/volumes.libsonnet';
 
 local version = 'nightly';
 {
-  PyTorchTest:: common.PyTorchTest {
-    frameworkPrefix: 'pt-%s' % version,
+  local Nightly = {
+    frameworkPrefix: 'pt-nightly',
```
JackCaoG, line 24: `version` was set to `nightly` above. What's difference between that and harding it to 'pytorch-nightly'>

tests/pytorch/nightly/mnist-3-7.libsonnet
```diff
@@ -1,66 +0,0 @@
-// Copyright 2020 Google LLC
```
JackCaoG, line 1: why is this config deleted?

will-cromar: Previously, all tests except for this one (and some of the CUDA tests) were running on the pytorch XLA `nightly` image with Python 3.6. This PR changes the image for all of the tests to the`nightly_3.7` Docker image that `mnist-3-7` was already using, making is obsolete.

tests/pytorch/nightly/common.libsonnet
will-cromar: I think the `version` var was added to make it easier to substitute in a release version for `frameworkPrefix`, `softwareVersion`, and `imageTag`. That didn't really work out since these tags all have subtly different formats; see `common.libsonnet` for r1.10:

https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/28ee676a0b56c1adf1fc1505b0c1ef214301b862/tests/pytorch/r1.10/common.libsonnet#L20-L41

To clarify, the net effect of my changes in this file is that the `imageTag` default changes from `nightly` to `nightly_3.7`. I also saw the opportunity to refactor and reduce repetition in this file. I double-checked the generated diffs to make sure the changes to `frameworkVersion` and `softwareVersion` has no effect.

PR #577, opened 2022-02-26
Run PyTorch setup steps on all TPU VM workers with `gcloud`

will-cromar:
This change makes the new `PyTorchTpuVmMixin` compatible with pod tests.

Replace fs_transformer v3-32 TPU Node `xla_dist` test with a TPU VM test. We still get coverage of the TPU Node + `xla_dist` case through ResNet:

https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/477803e1ca54c2834e9619d893ba766f01d4ebdd/tests/pytorch/nightly/resnet50-pod.libsonnet#L41

Both TPU VM transformer tests are currently broken due to libtpu issues (bug already filed internally). I ran test runs for v3-8 and v3-32 to confirm environment is set up correctly in device and pod cases. Also ran resnet on v3-8 to confirm there is no functional impact on non-transformer tests. Disabling v3-32 test until v3-8 starts passing.

Builds on work in #550

### DIFF
--- a/tests/experimental.libsonnet
+++ b/tests/experimental.libsonnet
@@ -118,6 +118,7 @@ local volumes = import 'templates/volumes.libsonnet';
                 exit $exit_code
               fi
 
+              echo ${zone} > /scripts/zone
               echo ${tpu_name} > /scripts/tpu_name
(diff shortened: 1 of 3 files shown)


JackCaoG (approved):
Thanks!

PR #597, opened 2022-03-31
Remove redundant PT/XLA tests.

will-cromar:
Delete r1.10 tests and r1.11 convergence tests since both releases are stable. Run r1.11 functional tests once per week (consistent with TF).

### DIFF
--- a/tests/pytorch/nightly/fs-transformer.libsonnet
+++ b/tests/pytorch/nightly/fs-transformer.libsonnet
@@ -203,10 +203,9 @@ local utils = import 'templates/utils.libsonnet';
   },
   configs: [
     transformer + v3_8 + functional_no_save + timeouts.Hours(1),
-    transformer + v3_8 + convergence + timeouts.Hours(25),
     transformer + v3_8 + convergence + timeouts.Hours(25) + tpuVm,
-    transformer + v3_8 + checkpoint_local + timeouts.Hours(2),
-    transformer + v3_8 + checkpoint_gcs + timeouts.Hours(2),
+    transformer + v2_8 + checkpoint_local + timeouts.Hours(2),
+    transformer + v2_8 + checkpoint_gcs + timeouts.Hours(2),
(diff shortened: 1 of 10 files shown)


tests/pytorch/nightly/fs-transformer.libsonnet
```diff
@@ -203,10 +206,9 @@ local utils = import 'templates/utils.libsonnet';
   },
   configs: [
     transformer + v3_8 + functional_no_save + timeouts.Hours(1),
-    transformer + v3_8 + convergence + timeouts.Hours(25),
     transformer + v3_8 + convergence + timeouts.Hours(25) + tpuVm,
-    transformer + v3_8 + checkpoint_local + timeouts.Hours(2),
-    transformer + v3_8 + checkpoint_gcs + timeouts.Hours(2),
+    transformer + v2_8 + checkpoint_local + timeouts.Hours(2),
```
JackCaoG, line 210: hmm, if checkpoint test is running using transformer, I am afraid it will OOM unless we update the config

tests/pytorch/nightly/resnet50-pod.libsonnet
```diff
@@ -87,14 +87,11 @@ local utils = import 'templates/utils.libsonnet';
...
   },
-  local v3_8 = {
-    accelerator: tpus.v3_8,
-  },
   local v3_32 = {
     accelerator: tpus.v3_32,
   },
   configs: [
-    resnet50_pod_func + v3_32 + functional,
+    resnet50_pod_func + v3_32 + functional + mixins.Experimental,
```
JackCaoG, line 94: what is this `mixins.Experimental`?

tests/pytorch/nightly/fs-transformer.libsonnet
will-cromar: Good point. Changing it back to v3-8

PR #598, opened 2022-03-31
Disable dlrm functional tests, except dlrm-mpdp-fwd

yeounoh:
Disable dlrm functional tests, except dlrm-mpdp-fwd, as it subsumes the other functional tests.

### DIFF
--- a/tests/pytorch/nightly/dlrm.libsonnet
+++ b/tests/pytorch/nightly/dlrm.libsonnet
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 local common = import 'common.libsonnet';
+local mixins = import 'templates/mixins.libsonnet';
 local timeouts = import 'templates/timeouts.libsonnet';
 local tpus = import 'templates/tpus.libsonnet';
 local utils = import 'templates/utils.libsonnet';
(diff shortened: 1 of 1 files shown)


tests/pytorch/nightly/dlrm.libsonnet
```diff
@@ -165,9 +165,9 @@ local utils = import 'templates/utils.libsonnet';
     accelerator: tpus.v3_8,
   },
   configs: [
-    dlrm + v3_8 + one_core + timeouts.Hours(3),
-    dlrm + v3_8 + seq_fwd + timeouts.Hours(3),
-    dlrm + v3_8 + mp_fwd + timeouts.Hours(3),
+    // dlrm + v3_8 + one_core + timeouts.Hours(3),
```
will-cromar, line 168: If you think it's worth keeping this config around, can you take them off the schedule with [`mixins.Experimental`](https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/4397c3976074fe29ec8c506a4e30343f9d40fb55/templates/mixins.libsonnet#L33)? This will still leave them available to run manually on an ad-hoc basis

If you don't think we need these configs anymore, please remove these lines and the corresponding mixins altogether.

yeounoh: I see, let's go with `mixins.Experimental` for future reference.

yeounoh: Thanks @will-cromar , so I assume that the `functional::schedule` will be overwritten by `::Experimental::schedule` ?

will-cromar: Yeah, that's right

will-cromar (approved):

PR #603, opened 2022-04-05
refact dlrm

miaoshasha:
tested by sending tests to gke

### DIFF
--- a/tests/pytorch/nightly/dlrm.libsonnet
+++ b/tests/pytorch/nightly/dlrm.libsonnet
@@ -19,46 +19,53 @@ local tpus = import 'templates/tpus.libsonnet';
 local utils = import 'templates/utils.libsonnet';
 
 {
-  local command_common = |||
-    pip install onnx
-    git clone --recursive https://github.com/pytorch-tpu/examples.git
-    python examples/deps/dlrm/dlrm_tpu_runner.py \
-      --arch-sparse-feature-size=64 \
-      --arch-mlp-bot=512-512-64 \
-      --arch-mlp-top=1024-1024-1024-1 \
-      --arch-interaction-op=dot \
-      --lr-num-warmup-steps 10 \
-      --lr-decay-start-step 10 \
-      --num-batches=1000 \
-      --data-generation="random" \
-      --numpy-rand-seed=727 \
-      --print-freq 100 \
-      --num-indices-per-lookup=100 \
-      --use-tpu \
-      --metrics-debug \
-      --num-indices-per-lookup-fixed \
-  |||,
-  local dlrm = common.PyTorchTest {
-    modelName: 'dlrm',
+  local dlrm = {
+    local config = self,
 
+    modelName: 'dlrm',
     volumeMap+: {
       datasets: common.datasetsVolume,
     },
-    podTemplate+:: {
-      spec+: {
-        containerMap+: {
-          train+: {
-            resources+: {
-              requests: {
-                cpu: '9.0',
-                memory: '30Gi',
-              },
-            },
-          },
+    paramsOverride:: {
+      scriptPath: 'tpu-examples/deps/dlrm/dlrm_tpu_runner.py',
+      miniBatchSize: 256,
+      archEmbeddingSize: '1000000-1000000',
+      tpuModelParallelGroupLen: 1,
+      tpuCores: 1,
+      ArchSparseFeatureSize: 64,
+      ArchMlpBot: '512-512-64',
+      ArchAlpTop: '1024-1024-1024-1',
+      NumIndicesPerLookup: 100,
+      trainCommand: [
+        'python3',
+        self.scriptPath,
+        '--arch-interaction-op=dot',
+        '--lr-num-warmup-steps=10',
+        '--lr-decay-start-step=10',
+        '--num-batches=1000',
+        '--data-generation="random"',
+        '--numpy-rand-seed=72',
+        '--print-freq=100',
+        '--use-tpu',
+        '--metrics-debug',
+        '--num-indices-per-lookup-fixed',
+        '--mini-batch-size=%d' % config.paramsOverride.miniBatchSize,
+        '--arch-embedding-size=%s' % config.paramsOverride.archEmbeddingSize,
+        '--tpu-model-parallel-group-len=%d' % config.paramsOverride.tpuModelParallelGroupLen,
+        '--tpu-cores=%d' % config.paramsOverride.tpuCores,
+      ],
+    },
+    cpu: '9.0',
+    memory: '30Gi',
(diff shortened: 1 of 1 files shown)


tests/pytorch/nightly/dlrm.libsonnet
```diff
@@ -82,94 +88,122 @@ local utils = import 'templates/utils.libsonnet';
...
+      ||| % [
+        utils.toCommandString(self.paramsOverride.trainCommand),
+      ]
     ),
   },
   local v3_8 = {
     accelerator: tpus.v3_8,
   },
   configs: [
-    dlrm + v3_8 + one_core + timeouts.Hours(3) + mixins.Experimental,
```
will-cromar, line 169: Can you mark these experimental per #598?

tests/pytorch/nightly/dlrm.libsonnet
```diff
@@ -19,42 +19,48 @@ local tpus = import 'templates/tpus.libsonnet';
...
-      --numpy-rand-seed=727 \
-      --print-freq 100 \
-      --num-indices-per-lookup=100 \
-      --use-tpu \
-      --metrics-debug \
-      --num-indices-per-lookup-fixed \
-  |||,
-  local dlrm = common.PyTorchTest {
-    modelName: 'dlrm',
+  local dlrm = {
```
will-cromar, line 22: Can you change this line back to `local dlrm = common.PyTorchTest {` and remove `common.PyTorchTest + ` from the configs below? If all of the tests use `common.PyTorchTest` as a base (which should always be the case, except for TPU Node Pod tests), it's better to include it in `dlrm` instead.

tests/pytorch/nightly/dlrm.libsonnet
```diff
@@ -82,94 +88,122 @@ local utils = import 'templates/utils.libsonnet';
...
+        '--learning-rate=0.1',
+        '--print-freq=1024',
+        '--no-save',
+        '--max-epoch=1',
+      ],
+    },
     command: utils.scriptCommand(
       |||
+        set +e
         apt-get install -y bc
```
will-cromar, line 183: Can you include these setup steps in `tpuVmExtraSetup` instead? Example:

https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/fef030dc06ecb685639230f05e2c317add1cb563/tests/pytorch/nightly/fs-transformer.libsonnet#L188-L196

tests/pytorch/nightly/dlrm.libsonnet
```diff
@@ -19,42 +19,48 @@ local tpus = import 'templates/tpus.libsonnet';
...
-              },
-            },
-          },
+    paramsOverride:: {
+      scriptPath: 'tpu-examples/deps/dlrm/dlrm_tpu_runner.py',
+      miniBatchSize: 256,
+      archEmbeddingSize: '1000000-1000000',
+      tpuModelParallelGroupLen: 1,
+      tpuCores: 1,
+      ArchSparseFeatureSize: 64,
```
will-cromar, line 35: Where are `ArchSparseFeatureSize`, `ArchMlpBot`, etc. used? I don't see them substituted anywhere in the `trainCommand`

also nit: use `camelCase` for these fields

tests/pytorch/nightly/dlrm.libsonnet
```diff
@@ -19,42 +19,48 @@ local tpus = import 'templates/tpus.libsonnet';
...
+        '--num-indices-per-lookup-fixed',
+        '--mini-batch-size=%d' % config.paramsOverride.miniBatchSize,
+        '--arch-embedding-size=%s' % config.paramsOverride.archEmbeddingSize,
+        '--tpu-model-parallel-group-len=%d' % config.paramsOverride.tpuModelParallelGroupLen,
+        '--tpu-cores=%d' % config.paramsOverride.tpuCores,
+      ],
+    },
+    cpu: '9.0',
+    memory: '30Gi',
+    metricConfig+: {
```
will-cromar, line 60: Do we need to disable TB assertions for this test? If not, please remove this `metricConfig` override

tests/pytorch/nightly/dlrm.libsonnet
```diff
@@ -82,94 +88,122 @@ local utils = import 'templates/utils.libsonnet';
...
       |||
+        set +e
         apt-get install -y bc
         pip install onnx
         git clone --recursive https://github.com/pytorch-tpu/examples.git
-        python examples/deps/dlrm/dlrm_tpu_runner.py \
-          %(convergence_common)s
-      ||| % convergence_common
+        %s 
+        --raw-data-file=/datasets/criteo-kaggle-mm/train.txt \
```
will-cromar, line 187: It looks like you already added these three flags to `trainCommand`, so these don't need to be here

will-cromar (changes requested):
Thanks for taking on this test. It's already looking better!

Can you also add a more detailed PR title and description explaining the changes and the context?

tests/pytorch/nightly/dlrm.libsonnet
miaoshasha: done!

tests/pytorch/nightly/dlrm.libsonnet
miaoshasha: makes sense

tests/pytorch/nightly/dlrm.libsonnet
miaoshasha: test dlrm tests are all 2vm tests, I can definitely move the four lines into a separate definition, but it won't be used elsewhere except this single case here. I'm keeping it as is if no other reasons.

tests/pytorch/nightly/dlrm.libsonnet
miaoshasha: thanks for the catch! changed to camelCase! these are substituted in criteo_kaggle.

tests/pytorch/nightly/dlrm.libsonnet
miaoshasha: right, btw what is TB assertions?

will-cromar: TensorBoard assertions. i.e. test expectations based out TensorBoard metrics

tests/pytorch/nightly/dlrm.libsonnet
will-cromar: I see, I missed that these were 2vm. You shouldn't need to clone the PyTorch examples for 2vm because this repository is already included in the PyTorch XLA docker image: https://github.com/pytorch/xla/blob/e8353135e0e057df495653a7bc8ac2c00ce60b82/docker/Dockerfile#L49

PR #596, opened 2022-03-31
Increase timeout for nightly python-ops tests

wonjoolee95:
Increasing timeout for `python-ops` tests as they have been failing for a while due to timeouts.

PR #605, opened 2022-04-06
Update timeout for python-op tests again

wonjoolee95:
The previous timeout increase PR https://github.com/GoogleCloudPlatform/ml-testing-accelerators/pull/596 was not enough, as tests are still timing out. 

cc @miladm @will-cromar


will-cromar (approved):

will-cromar:
Let me know if this needs to be re-opened.

PR #614, opened 2022-04-13
use bfloat for convergence test to take 256x64 512x32 input shape

miaoshasha:
bfloat needed for such shape


miladm (approved):
Thanks @miaoshasha 

LGTM once tests pass.

PR #618, opened 2022-04-21
Update actions-x/commit in `generate.yml` workflow

will-cromar:
Upgrading actions-x/commit to v5 per their [README](https://github.com/actions-x/commit/blob/1929419481a5b78d564c9f71e99309e7fccacf8b/README.md?plain=1#L3)

Fixes broken CI workflow


will-cromar:
Test run: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/runs/6116783209?check_suite_focus=true

allenwang28 (approved):

PR #621, opened 2022-04-21
Flax nightly tests are not necessary because we already have jax nigh…

sshahrokhi:
…tly and Flax team only wants latests tested,

Keeping the targets file here, if we wanted to add tests in the future.


will-cromar:
If these are helpful to run on a one-off basis, it's better to set the `schedule` field to `null` for these tests. The best way to do this is to add a shared mixin to `common`. That's what we do for old TF releases:

https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/0b38c5741597a0b935b86ec49f7c5b001e036a63/tests/tensorflow/r2.5/common.libsonnet#L52

If you don't plan to run them ever, it's better to remove this directory entirely rather than have an empty `targets.libsonnet`. We always have the change history (not to mention the `latest` directory) if we need to bring these configs back.

sshahrokhi:
> If these are helpful to run on a one-off basis, it's better to set the `schedule` field to `null` for these tests. The best way to do this is to add a shared mixin to `common`. That's what we do for old TF releases:
> 
> https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/0b38c5741597a0b935b86ec49f7c5b001e036a63/tests/tensorflow/r2.5/common.libsonnet#L52
> 
> If you don't plan to run them ever, it's better to remove this directory entirely rather than have an empty `targets.libsonnet`. We always have the change history (not to mention the `latest` directory) if we need to bring these configs back.

I removed the targets, because I really don't think we ever need to run these tests while we have the jax tests running.

will-cromar:
Sounds good. I would go a step further and remove the entire `jax/nightly` directory if none of the targets inside (`imagenet.libsonnet`, `mnist.libsonnet`, etc) will get run

sshahrokhi:
Currently jax nightly tests are not in the nightly folder, (they are jax unit tests) so for now I removed the whole nightly folder because it only had flax nightly tests.

will-cromar (approved):

PR #624, opened 2022-04-28
Change the frequency of old TF versions and put TF-2.9.0 at highest priority. Add TF-2.9.0 to targets for launching them as workloads.

aman2930:
- Add the 2.9 tests to targets for creating final workloads in the test project.
- Completely stopped r2.7 tests.
- Reduced the frequency of r2.8 tests and stopped its convergence tests.
- Set the time for r2.9 functional tests to 9 PM and convergence tests to 3 hours later @12 AM.
- Removed some old hanging r2.4 and r2.5 tests.


chandrasekhard2 (approved):

will-cromar (approved):
LGTM except for build issues.

tests/tensorflow/r2.9/common.libsonnet
```diff
@@ -76,7 +73,9 @@ local mixins = import 'templates/mixins.libsonnet';
   RunNightly:: {
     schedule: functional_schedule,
   },
+  // Running functional tests at Midnight PST Daily.
```
will-cromar, line 76: Do you want to run convergence tests every day?

aman2930: Yes. For now. We are already late this release cycle. Tensorflow is close to their release, and we haven't tested even RC0 yet. So want to start testing daily. Once we get enough confidence, we will reduce the frequency of the tests.

PR #627, opened 2022-05-03
fix transformer and bert paths

chandrasekhard2:


tests/tensorflow/r2.7/retinanet.libsonnet
```diff
@@ -44,7 +44,7 @@ local tpus = import 'templates/tpus.libsonnet';
     },
     command: [
       'python3',
-      'official/vision/detection/main.py',
```
will-cromar, line 47: The legacy folder doesn't exist for 2.7 and below: https://github.com/tensorflow/models/tree/r2.7.0/official

will-cromar (changes requested):

will-cromar (approved):

PR #629, opened 2022-05-03
Set --metrics_debug=false for PT MNIST GPU test

will-cromar:


JackCaoG (approved):

will-cromar:
This test currently fails in `xla::XrtComputationClient::GetMetrics()`. Skip call to `met.metrics_report`

PR #630, opened 2022-05-05
Merge PyTorch ResNet50 tests

will-cromar:
Merge together all of the different test configurations for ResNet on PyTorch. As a side effect, this means the v3-8 TPU VM convergence test will actually run for 90 epochs.


JackCaoG (approved):
Thanks!

PR #637, opened 2022-05-16
Disable PT/XLA metrics report for MNIST and ResNet on GPU

will-cromar:


will-cromar:
Tested manually with MNIST

JackCaoG (approved):
Thanks!

PR #641, opened 2022-05-20
Add TF-Vision RetinaNet test

A9isha:
Add TF `vision-retinanet` test

### DIFF
--- a/tests/tensorflow/nightly/vision-retinanet.libsonnet
+++ b/tests/tensorflow/nightly/vision-retinanet.libsonnet
@@ -20,30 +20,27 @@ local tpus = import 'templates/tpus.libsonnet';
 local utils = import 'templates/utils.libsonnet';
 
 {
-  local retinanet = common.TfVisionTest {
+  local coco = {
+    scriptConfig+: {
+      trainFilePattern: '$(COCO_DIR)/train*',
+      evalFilePattern: '$(COCO_DIR)/val*',
+    },
+  },
+  local retinanet = common.TfVisionTest + coco {
     modelName: 'vision-retinanet',
     scriptConfig+: {
       experiment: 'retinanet_resnetfpn_coco',
-    },
-    PreemptibleTpu:: {
-      tpuSettings+: {
-        preemptible: false,
+      paramsOverride+: {
+        task+: {
+          annotation_file: '$(COCO_DIR)/instances_val2017.json',
+        },
       },
     },
   },
   local functional = common.Functional {
     scriptConfig+: {
       paramsOverride+: {
-        task: {
-          annotation_file: '$(COCO_DIR)/instances_val2017.json',
-	   train_data: {
-             input_path: '$(COCO_DIR)/train*',
-          },
-	   validation_data: {
-             input_path: '$(COCO_DIR)/val*',
-          },
-        },
-        trainer: {
+        trainer+: {
           train_steps: 200,
           validation_interval: 200,
           validation_steps: 100,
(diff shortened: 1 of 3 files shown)


tests/tensorflow/nightly/vision-retinanet.libsonnet
```diff
@@ -0,0 +1,113 @@
...
+      },
+    },
+  },
+  local functional_v2_8 = common.Functional {
+    scriptConfig+: {
+      paramsOverride+: {
+        task: {
+          annotation_file: '$(COCO_DIR)/instances_val2017.json',
+          train_data: {
+	    global_batch_size: 64,
```
will-cromar: Add this to `v2_8` instead:

```
      paramsOverride+: {
        task: {
          train_data: {
	    global_batch_size: 64,
          },
        },
      },
```

This way, the batch size change takes effect for both the functional and convergence tests, and you don't have to have two copies of `functional`.

tests/tensorflow/nightly/vision-retinanet.libsonnet
```diff
@@ -0,0 +1,113 @@
...
+    PreemptibleTpu:: {
+      tpuSettings+: {
+        preemptible: false,
+      },
+    },
+  },
+  local functional = common.Functional {
+    scriptConfig+: {
+      paramsOverride+: {
+        task: {
```
will-cromar, line 37: Move these common fields to `retinanet` instead:

```
      paramsOverride+: {
        task: {
          annotation_file: '$(COCO_DIR)/instances_val2017.json',
	  train_data: {
            input_path: '$(COCO_DIR)/train*',
          },
	  validation_data: {
            input_path: '$(COCO_DIR)/val*',
          },
        },
```

The input data is common to both `functional` and `convergence`, but right now the convergence test doesn't point to any input data.

tests/tensorflow/nightly/vision-retinanet.libsonnet
```diff
@@ -0,0 +1,113 @@
...
+local tpus = import 'templates/tpus.libsonnet';
+local utils = import 'templates/utils.libsonnet';
+
+{
+  local retinanet = common.TfVisionTest + {
+    modelName: 'vision-retinanet',
+    scriptConfig+: {
+      experiment: 'retinanet_resnetfpn_coco',
+    },
+    PreemptibleTpu:: {
```
will-cromar, line 28: Remove this `PreemptibleTpu` block. It probably doesn't do what you want it to, and we want to control TPU preemptibility in common templates.

will-cromar (changes requested):

tests/tensorflow/nightly/vision-retinanet.libsonnet
```diff
@@ -0,0 +1,107 @@
+// Copyright 2021 Google LLC
```
will-cromar, line 1: 2022

tests/tensorflow/nightly/vision-retinanet.libsonnet
```diff
@@ -0,0 +1,107 @@
...
+        },
+      },
+    },
+  },
+  local val_32 = {
+    scriptConfig+: {
+      paramsOverride+: {
+        task+: {
+          validation_data+: {
+            global_batch_size: 32,
```
will-cromar, line 69: I would create a common TPU subclass where you compute the validation batch size according to the number of replicas (i.e. `config.accelerator.replicas`). See this example:

https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/7dec2008823f97ec67a4ab444a8975d4aa2e1ad2/tests/tensorflow/r2.8/maskrcnn.libsonnet#L91-L93

will-cromar (changes requested):
Otherwise LGTM

will-cromar:
Can you also rename this file `vision-coco` to align with `vision-imagenet`? We may want to add more tests that use the same dataset

tests/tensorflow/nightly/vision-retinanet.libsonnet
A9isha: done

will-cromar (approved):
LGTM. Great work!

PR #644, opened 2022-05-25
Add PjRt tests for MNIST and ResNet50

will-cromar:
Also add comparable fake_data test for ResNet50 on XRT


JackCaoG:
my understanding s resnet with PJRT still oom right now, is that the case?

will-cromar:
I still haven't observed an OOM with fake data and full epochs

JackCaoG (approved):
Thanks!

PR #645, opened 2022-05-26
add tfvision maskrcnn tests

khshah6:
Add TF vision-maskrcnn test

Rebased off #641 and added maskrcnn tests to same file

oneshot test results at gs://xl-ml-test-us-central1/output/tf-nightly/vision-maskrcnn/func/v3-8/tf-nightly-vision-maskrcnn-func-v3-8-7w8jp/ckpt-400


tests/tensorflow/nightly/targets.jsonnet
```diff
@@ -27,6 +27,7 @@ local shapemask = import 'shapemask.libsonnet';
 local transformer_translate = import 'transformer-translate.libsonnet';
 local vision_coco = import 'vision-coco.libsonnet';
 local vision_imagenet = import 'vision-imagenet.libsonnet';
+local vision_coco = import 'vision-coco.libsonnet';
```
will-cromar, line 30: This is probably leftover from a merge. `vision_coco` is already defined on line 28

tests/tensorflow/nightly/vision-coco.libsonnet
```diff
@@ -49,6 +49,17 @@ local utils = import 'templates/utils.libsonnet';
       },
     },
   },
+  local maskrcnn = common.TfVisionTest + coco {
+    modelName: 'vision-maskrcnn',
+    scriptConfig+: {
+      experiment: 'maskrcnn_resnetfpn_coco',
+      paramsOverride+: {
+        task+: {
+          annotation_file: '$(COCO_DIR)/instances_val2017.json',
```
will-cromar, line 58: Is `task.annotation_file` always required for coco? If so, this override should move to `coco`. (same for `retinanet`, even though it wasn't added in this PR)

will-cromar (changes requested):
Mostly LGTM aside from two nits.

Were you able to do a successful test run?

tests/tensorflow/nightly/vision-coco.libsonnet
khshah6: Yup looks like it, will do

khshah6:
Currently working on test run

tests/tensorflow/nightly/vision-coco.libsonnet
```diff
@@ -75,6 +86,15 @@ local utils = import 'templates/utils.libsonnet';
   },
   local v3_8 = tpu_common {
     accelerator: tpus.v3_8,
+    scriptConfig+: {
+      paramsOverride+: {
+        task+: {
+          train_data+: {
+            global_batch_size: 64,
```
will-cromar, line 93: Does the default batch size not work for v3-8? If there's a constant train batch size per core/replica, it's better to compute the global batch size in the base object (`maskrcnn` in this case). Otherwise, we have to go through and set the global batch size manually for each case. See the validation batch size in `tpu_common` for an example.

tests/tensorflow/nightly/vision-coco.libsonnet
```diff
@@ -42,9 +47,15 @@ local utils = import 'templates/utils.libsonnet';
...
-          annotation_file: '$(COCO_DIR)/instances_val2017.json',
+    },
+  },
+  local maskrcnn = common.TfVisionTest + coco {
+    modelName: 'vision-maskrcnn',
+    scriptConfig+: {
+      experiment: 'maskrcnn_resnetfpn_coco',
+      paramsOverride: {
+        trainer: {
+          train_steps: 400,
```
will-cromar, line 58: FYI, this is also setting the convergence train steps to 400, and the functional test still gets overridden to 200 steps by `functional`

will-cromar (changes requested):

tests/tensorflow/nightly/vision-coco.libsonnet
khshah6: ah ok I see

tests/tensorflow/nightly/vision-coco.libsonnet
```diff
@@ -74,7 +69,7 @@ local utils = import 'templates/utils.libsonnet';
   local convergence = common.Convergence,
   local v2_8 = {
     accelerator: tpus.v2_8,
-    scriptConfig+: {
+    scriptConfig+: {a
```
will-cromar, line 72: This doesn't look intentional :)

will-cromar (approved):
LGTM. Thanks!

PR #648, opened 2022-05-26
Add PjRt ResNet50 convergence test.

will-cromar:


will-cromar:
Also disabled summaries for the convergence test. We increased the number of epochs, which threw historical values out of bounds. Will re-enable metrics once we build up a new history.

JackCaoG (approved):
Thanks!

PR #651, opened 2022-05-31
Add 1.12 tests

wonjoolee95:
Add 1.12 tests

### DIFF
--- a/tests/pytorch/r1.12/common.libsonnet
+++ b/tests/pytorch/r1.12/common.libsonnet
@@ -1,0 +1,66 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+local common = import '../common.libsonnet';
+local experimental = import '../experimental.libsonnet';
+local mixins = import 'templates/mixins.libsonnet';
+local utils = import 'templates/utils.libsonnet';
+local volumes = import 'templates/volumes.libsonnet';
+
+{
+  local r1_12 = {
+    frameworkPrefix: 'pt-r1.12',
+    tpuSettings+: {
+      softwareVersion: 'pytorch-1.12',
+    },
+    imageTag: 'r1.12',
+  },
+  PyTorchTest:: common.PyTorchTest + r1_12,
+  PyTorchXlaDistPodTest:: common.PyTorchXlaDistPodTest + r1_12,
+  PyTorchGkePodTest:: common.PyTorchGkePodTest + r1_12,
+  Functional:: mixins.Functional {
+    schedule: '0 7 * * *',
+    tpuSettings+: {
+      preemptible: false,
+    },
+  },
+  Convergence:: mixins.Convergence,
+  PyTorchTpuVmMixin:: experimental.PyTorchTpuVmMixin {
+    tpuSettings+: {
+      softwareVersion: 'tpu-vm-base',
+      tpuVmPytorchSetup: |||
+        sudo pip3 uninstall --yes torch torch_xla torchvision numpy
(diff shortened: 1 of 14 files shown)


JackCaoG:
I am assuming this is just copy from 1.11 and change it to 1.12? Is there a reason that it is still a draft? Do we need to dry run this before merging this?

wonjoolee95:
> I am assuming this is just copy from 1.11 and change it to 1.12? Is there a reason that it is still a draft? Do we need to dry run this before merging this?

Yes, this is just a copy from 1.11 to 1.12 with some changes to 1.12 branch. We should be good to merge this, this won't affect any of the existing tests.

JackCaoG (approved):
Thanks! Let's try to get some run this weekend.

will-cromar (changes requested):
I would copy from nightly to 1.12 instead of 1.11 to 1.12. 1.11 had quite a few broken test configs that have been fixed recently. Just remove the configs that shouldn't block the release (especially PjRt).

wonjoolee95:
> I would copy from nightly to 1.12 instead of 1.11 to 1.12. 1.11 had quite a few broken test configs that have been fixed recently. Just remove the configs that shouldn't block the release (especially PjRt).

Thanks for pointing that out, @will-cromar. I updated the PR to copy from nightly tests instead. Also removed tests for `pjrt`. I'll follow-up with separate PRs on removing tests that we don't need for r1.12 release.

tests/pytorch/r1.12/common.libsonnet
```diff
@@ -0,0 +1,66 @@
...
+    },
+  },
+  Convergence:: mixins.Convergence,
+  PyTorchTpuVmMixin:: experimental.PyTorchTpuVmMixin {
+    tpuSettings+: {
+      softwareVersion: 'tpu-vm-base',
+      tpuVmPytorchSetup: |||
+        sudo pip3 uninstall --yes torch torch_xla torchvision numpy
+        sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch-1.12-cp38-cp38-linux_x86_64.whl
+        sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch_xla-1.12-cp38-cp38-linux_x86_64.whl
```
will-cromar, line 45: Do we need to install libtpu too?

tests/pytorch/r1.12/common.libsonnet
```diff
@@ -0,0 +1,66 @@
...
+        sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torchvision-1.12-cp38-cp38-linux_x86_64.whl
+        sudo pip3 install numpy
+        sudo pip3 install mkl mkl-include cloud-tpu-client
+        sudo apt-get -y update
+        sudo apt-get install -y libomp5
+        # No need to check out the PyTorch repository, but check out PT/XLA at
+        # pytorch/xla anyway
+        mkdir pytorch
+        cd pytorch
+        git clone https://github.com/pytorch/xla.git -b r1.11
```
will-cromar, line 55: 1.12

will-cromar (approved):
Thanks @wonjoolee95! Just a couple of comments

tests/pytorch/r1.12/common.libsonnet
wonjoolee95: Absolutely, good catch. Added, `sudo pip3 install https://storage.googleapis.com/cloud-tpu-tpuvm-artifacts/wheels/libtpu-nightly/libtpu_nightly-0.1.dev20220518-py3-none-any.whl`.

wonjoolee95:
Thanks! I don't seem to have merging rights. Please merge this when you get the time, thanks 😄, @will-cromar!

will-cromar (commented):
I just checked the diff and there aren't any new files generated. You'll need to add 1.12 here: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/pytorch/targets.jsonnet

tests/pytorch/targets.jsonnet
```diff
@@ -15,9 +15,11 @@
 local nightly = import 'nightly/targets.jsonnet';
 local r1_10 = import 'r1.10/targets.jsonnet';
 local r1_11 = import 'r1.11/targets.jsonnet';
+local r1_11 = import 'r1.12/targets.jsonnet';
```
will-cromar, line 18: r1_12

wonjoolee95: Thanks lol, definitely not used to editing in vim

PR #663, opened 2022-06-14
Verify fairseq-generate being installed

JackCaoG:
For some reason `fairseq-generate` is not available after `fairseq` installation. I was not able to repo locally on the xlml test docker. Adding this and manually trigger the test to help the debug.

error is `/bin/bash: line 46: fairseq-generate: command not found` after 25 epoches of training is completed.

### DIFF
--- a/tests/pytorch/nightly/fs-transformer.libsonnet
+++ b/tests/pytorch/nightly/fs-transformer.libsonnet
@@ -159,9 +159,7 @@ local utils = import 'templates/utils.libsonnet';
     command: utils.scriptCommand(
       |||
         export XLA_USE_BF16=1
-        pip install --editable tpu-examples/deps/fairseq
-        which fairseq-generate
-        fairseq-generate -h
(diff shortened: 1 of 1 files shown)


will-cromar:
It's currently the TPU VM version of this test that is hitting the error, which does not use the docker image. To reproduce, you can try running these setup steps on TPU VM and figure out why `fairseq-generate` is not on the PATH:

https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/4bb7bdf01de7c631ccc983efd0f843138346c7b1/k8s/us-central1/gen/pt-nightly-fs-transformer-conv-v3-8-1vm.yaml#L135

JackCaoG:
Ah, I didn't know TPUVM does not use docker.

JackCaoG:
I verified that `sudo pip install` solved the issue, it is a permission thing.

will-cromar (approved):

JackCaoG:
Unless we will run this test in the TPUVM docker, I think using `sudo` should be fine

PR #664, opened 2022-06-14
Refactor roberta-pre tests

A9isha:


tests/pytorch/nightly/roberta-pre.libsonnet
```diff
@@ -19,67 +19,68 @@ local tpus = import 'templates/tpus.libsonnet';
...
-                'ephemeral-storage': '10Gi',
-              },
-            },
-          },
-        },
-      },
-    },
+    cpu: '9.0',
+    memory: '30Gi',
+    'ephemeral-storage': '10Gi',
```
will-cromar, line 83: The base class doesn't use this field like it does `cpu` and `memory`, so this line has no effect. I don't object to removing the ephemeral storage requirement and seeing what happens though.

will-cromar (changes requested):

tests/pytorch/nightly/roberta-pre.libsonnet
A9isha: yes - that's correct - the git diff showed the ephemeral-storage setting removed even if the source code here had it - I have removed it from the source code now.

will-cromar (approved):
Great work! This file looks much better now

PR #665, opened 2022-06-14
Remove version specify from jax install command.

skye:
Now that we have vanilla base images without jax preinstalled, we
don't need this version hack to make pip install the latest jax
version.

I'll update the JAX README and other documentation once we test this
actually works.


skye:
Can you merge this as well? I don't have permissions (or alternatively, give me permissions :))

PR #669, opened 2022-06-16
Lower accuracy assertion for ResNet50.

will-cromar:


tests/pytorch/nightly/resnet50-mp.libsonnet
```diff
@@ -65,7 +65,7 @@ local tpus = import 'templates/tpus.libsonnet';
               FINAL: {
                 fixed_value: {
                   comparison: 'GREATER',
-                  value: 75.0,
+                  value: 74.0,
```
JackCaoG, line 68: Does this change both pod and donut?

will-cromar: Yes. Do we want to keep it at 75% for donut?

JackCaoG: ideally yes.. This way when donut regress we can catch that.

will-cromar: Done. Checked the diff and this only affects the pod version of the test.

JackCaoG (approved):

wonjoolee95 (approved):

PR #679, opened 2022-06-23
Update base template to make writing simple tests more simple.

will-cromar:
Fix some usability issues I saw when rewriting our tutorials.

- Remove hard-coded requirement for a `gcs-buckets` `ConfigMap`.
- Add `entrypoint` override option.
- Update `tests/common.libsonnet` so our generated files stay the same.


will-cromar:
Confirmed there's no impact to generated files.

PR #683, opened 2022-06-29
Update tensorflow test nlp-wmt.libsonnet to fix timeout issue

aireenmei:
The same test is timeout for both tf-nightly and tf-2.9.1
Compared with the test in tf2.8, I updated train steps and batch size.
Test pass on tf-nightly-nlp-wmt-transformer-conv-v2-8

### DIFF
--- a/tests/tensorflow/r2.9/nlp-wmt.libsonnet
+++ b/tests/tensorflow/r2.9/nlp-wmt.libsonnet
@@ -12,12 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-local experimental = import '../experimental.libsonnet';
 local common = import 'common.libsonnet';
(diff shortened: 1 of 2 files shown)


tests/tensorflow/r2.9/nlp-wmt.libsonnet
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+local experimental = import '../experimental.libsonnet';
```
gagika, line 15: Do you need this?

aireenmei: Thanks for pointing it out! I then removed the unused imports in the new commit

gagika (approved):

PR #684, opened 2022-06-30
Remove conv tests for r1.12

wonjoolee95:
Remove conv tests for r1.12


tests/pytorch/r1.12/common.libsonnet
```diff
@@ -30,12 +30,11 @@ local volumes = import 'templates/volumes.libsonnet';
   PyTorchXlaDistPodTest:: common.PyTorchXlaDistPodTest + r1_12,
   PyTorchGkePodTest:: common.PyTorchGkePodTest + r1_12,
   Functional:: mixins.Functional {
-    schedule: '0 7 * * *',
+    schedule: '0 7 * * 1',
```
wonjoolee95, line 33: @will-cromar, is this the correct way to make this run only on Monday's?

JackCaoG (approved):

tests/pytorch/r1.12/common.libsonnet
```diff
@@ -30,12 +30,11 @@ local volumes = import 'templates/volumes.libsonnet';
...
     tpuSettings+: {
       preemptible: false,
     },
   },
-  Convergence:: mixins.Convergence,
```
will-cromar, line 38: If you set the schedule field to null, then we won't generate any of the conv tests without having to remove them from the repository yet:

```
Convergence:: mixins.Convergence {
  schedule: null
}
```

will-cromar (changes requested):

tests/pytorch/r1.12/common.libsonnet
wonjoolee95: This is very helpful! Updated.

will-cromar (approved):

PR #687, opened 2022-07-20
Move TF Nightly Keras tests to TPU VM

ssusie:
Converting existing Keras API tests that don't fail to run on 1vm


tests/tensorflow/nightly/common.libsonnet
```diff
@@ -26,6 +27,20 @@ local mixins = import 'templates/mixins.libsonnet';
     },
     imageTag: 'nightly',
   },
+  // Setting the version for TPU VM.
+  tpuVm:: experimental.TensorFlowTpuVmMixin {
+    local config = self,
+    tpuSettings+: {
```
will-cromar, line 33: In general, this directory should only have nightly tests. If these tests only work on 2.9, you can make the changes here instead: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/tensorflow/r2.9/keras-api.libsonnet

will-cromar (changes requested):

tests/tensorflow/nightly/common.libsonnet
```diff
@@ -26,6 +26,7 @@ local mixins = import 'templates/mixins.libsonnet';
     },
     imageTag: 'nightly',
   },
+
```
will-cromar, line 29: please revert this line

will-cromar (approved):

PR #690, opened 2022-07-29
refactoring to use gcloud ssh worker=all

ssusie:
This is for JAXPodTest only

### DIFF
--- a/tests/jax/common.libsonnet
+++ b/tests/jax/common.libsonnet
@@ -67,29 +67,8 @@ local tpus = import 'templates/tpus.libsonnet';
       |||
         set -x
         set -u
-        ssh -i scripts/id_rsa -o StrictHostKeyChecking=no \
-            xl-ml-test@$(cat /scripts/tpu_ip) << 'TEST_SCRIPT_EOF'
-          %s
-        TEST_SCRIPT_EOF
-        exit_code=$?
-        bash /scripts/cleanup.sh
-        exit $exit_code
-      ||| % config.testScript,
-    ],
-  },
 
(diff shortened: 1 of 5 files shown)


tests/jax/common.libsonnet
```diff
@@ -80,56 +89,6 @@ local tpus = import 'templates/tpus.libsonnet';
 
   JaxPodTest:: self.JaxTest {
```
will-cromar, line 90: Since `JaxPodTest` is a copy of `JaxTest`, you can remove `JaxPodTest` and replace it with `JaxTest` wherever it's used

I believe you can do the same for `PodFlaxLatest` (feel free to do this in a follow-up if it's not trivial): 

https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/8d5d7e9dcfe96b4b4d011ef0d2e5eafeafc0cff3/tests/jax/latest/common.libsonnet#L64

will-cromar (changes requested):
Thanks! This is much cleaner

will-cromar (approved):

sshahrokhi (commented):
Thanks for this it is much cleaner! 
Is there any difference now between PodFlaxLatest and runFlaxLatest? I skimmed and found no difference, if that's the case, maybe you could just remove the PodFlaxLatest part completely.

tests/jax/common.libsonnet
sshahrokhi: yeah it is true it seems.

ssusie:
Thanks @sshahrokhi for noticing that. Removed and replaced.

sshahrokhi (approved):

PR #695, opened 2022-08-15
Remove PyTorch testing image

will-cromar:
All our custom testing image did is override the entrypoint to set the XRT TPU config. Using the new `entrypoint` field to set the same variable.

Now we can directly use the official `gcr.io/tpu-pytorch/xla` images without building a parallel set of images.

The GPU examples image wasn't being used anyway.

We still do need the `pytorch-pods` image for TPU Node pod tests because it bundles some scripts we need. However, it's the same for every release, so I set the tag to `nightly` for all releases.


wonjoolee95 (approved):
Thanks! 👍

PR #703, opened 2022-08-30
Add eval metrics back to FS transformer

will-cromar:


darisoy (approved):
LGTM

will-cromar:
Leaving the old log parsing logic out. It will still be useful to have the final eval metrics in the logs, but we should find some better way to extract them to the metrics handler.

JackCaoG (approved):

PR #704, opened 2022-08-30
Match TPU Node build date to libtpu pin in PT/XLA

will-cromar:
See https://cloud.google.com/tpu/docs/version-switching#additional_software_options

Tested manually with MNIST on v2-8

### DIFF
--- a/tests/pytorch/nightly/common.libsonnet
+++ b/tests/pytorch/nightly/common.libsonnet
@@ -125,7 +125,10 @@ local volumes = import 'templates/volumes.libsonnet';
 
     podTemplate+:: {
       spec+: {
-        containerMap+: {
+        initContainerMap+:: {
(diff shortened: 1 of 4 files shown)


tests/pytorch/nightly/common.libsonnet
```diff
@@ -26,7 +26,55 @@ local volumes = import 'templates/volumes.libsonnet';
...
+              'python3',
+              '-c',
+              |||
+                import importlib_metadata
+                import os
+                import re
+
+                import cloud_tpu_client
+
+                requirements = importlib_metadata.requires('torch_xla')
```
JackCaoG, line 57: oh nice, didn't know we can do this.

JackCaoG (approved):
Thanks!

PR #705, opened 2022-08-31
Refactor pt nightly dlrm tests in order to later add 1vm tests

darisoy:


tests/pytorch/nightly/dlrm.libsonnet
```diff
@@ -19,45 +19,44 @@ local tpus = import 'templates/tpus.libsonnet';
...
-            },
-          },
-        },
-      },
+    paramsOverride:: {
+      scriptPath: 'tpu-examples/deps/dlrm/dlrm_tpu_runner.py',
+      miniBatchSize: 256,
+      archEmbeddingSize: '1000000-1000000',
+      tpuModelParallelGroupLen: 1,
+      tpuCores: 1,
```
will-cromar, line 34: This should default to `config.accelerator.numCores` so we get the correct number of cores for TPU v4: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/224065875e2505ddbd26bcfcc2c6ccdddb8d3ab8/templates/tpus.libsonnet#L25

Remove `tpuCores: 8` from `seq_fwd`, `mp_fwd`, etc too, since that will only work on v3

will-cromar (approved):
LGTM

PR #709, opened 2022-09-02
Add missing `python3` to fs-transformer conv test

will-cromar:
Currently fails with this error: `/bin/bash: line 42: fairseq-pjrt/generate.py: Permission denied`


JackCaoG (commented):
if it is a permission issue why adding a `python` would help? Shouldn't we add a sudo?

will-cromar:
It's getting permission denied because `generate.py` isn't marked executable, since it's meant to be invoked with python

JackCaoG (approved):

PR #724, opened 2022-10-10
Enable cross-account publish on event-publisher

jonb377:
1VM tests provisioned using QueuedResources are hosted in a separate project, so to minimize infrastructure duplication the event publisher should be able to publish cross-account directly to the main project's metrics topic. This requires that `--topic` accepts a full topic resource name.

For backwards compatibility, the `--topic` flag will continue to accept the shorthand topic name for topics within the same project as the publisher.


will-cromar (approved):
Thanks!

PR #730, opened 2022-10-20
Start directly using VM image in r1.13 xl ml tests as opposed to using a base image and install wheels.

vanbasten23:


tests/pytorch/r1.13/common.libsonnet
```diff
@@ -90,14 +90,10 @@ local volumes = import 'templates/volumes.libsonnet';
 
     tpuSettings+: {
       softwareVersion: if config.accelerator.version < 4 then
-        'tpu-vm-base'
+        'projects/cloud-tpu-v2-images-dev/global/images/tpu-vm-base-pt-1-13-0-20221019'
```
vanbasten23, line 93: Note to myself: seems equivalent to `projects/cloud-tpu-v2-images-dev/global/images/family/tpu-vm-base-pt-1-13-0`

vanbasten23: But we should use the image names that match the ones in the cdpush config. This is because the TPU API is separate from the GCE API. So the images are named differently. These are the names that get registered in the TPU API: cdpush/cloud-tpus/prod/config.gcl;l=801 and cdpush/cloud-tpus/prod/config.gcl;l=811. 

Despite that we can do
```
$ curl -X POST https://tpu.googleapis.com/v2alpha1/projects/tpu-prod-env-one-vm/locations/europe-west4-a/nodes?node_id=$USER -H "Authorization: Bearer $(gcloud auth print-access-token)" -H "Content-Type: application/json" -d '{accelerator_type: "v2-8", runtime_version: "tpu-vm-base", network_config: {enable_external_ips: true}, boot_disk: {source_image: "projects/cloud-tpu-v2-images-dev/global/images/tpu-vm-base-pt-1-13-0-20221019"}}'
```
, it's because boot_disk is a hidden API field but Normally users can only use runtime_version.

will-cromar (approved):

PR #734, opened 2022-10-26
Fix jax compilatioin cache test

yejingxin:
DETAILS:
Current computation has num of instruction = 5 which is below threshold=6, set threshold=0 to fix the test.

TESTED:
passed one-off test on
job.batch/jax-compilation-cache-test-func-v2-8-1vm-wmfq7

### DIFF
--- a/tests/jax/compilation-cache.libsonnet
+++ b/tests/jax/compilation-cache.libsonnet
@@ -48,7 +48,9 @@ local mixins = import 'templates/mixins.libsonnet';
       from jax import pmap, lax
       from jax._src.util import prod
       import numpy as np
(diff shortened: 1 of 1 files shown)


tests/jax/compilation-cache.libsonnet
```diff
@@ -48,7 +48,9 @@ local mixins = import 'templates/mixins.libsonnet';
       from jax import pmap, lax
       from jax._src.util import prod
       import numpy as np
+      from jax._src.config import config
```
skye, line 51: nit: move this two lines up

skye (approved):

PR #735, opened 2022-10-26
Remove Pytorch/xla r1.12 tests

vanbasten23:
Remove old r1.12 test because most of them are timing out due to low capacity.


JackCaoG (commented):
Can we also delete the 1.12 dir directly?

vanbasten23:
> Can we also delete the 1.12 dir directly?

Yes, I don't see a reason why not. Also directory r1.10 and r1.11 still exist. Do you want to delete them as well?

JackCaoG:
> > Can we also delete the 1.12 dir directly?
> 
> Yes, I don't see a reason why not. Also directory r1.10 and r1.11 still exist. Do you want to delete them as well?

yea, let's delete all of those unused dir.

JackCaoG (approved):

PR #739, opened 2022-11-04
Fix jax compilation cache test

yejingxin:
DETAILS:
Recent jax update threshold to be
jax_persistent_cache_min_compile_time_secs, set it to be zero to fix the test.

TESTED:
passed one-off test on
job_name=job.batch/jax-compilation-cache-test-func-v2-8-1vm-tkqhf


skye (approved):
Thanks!

PR #741, opened 2022-11-10
Replace `sklearn` with `scikit-learn`

will-cromar:
Fixes `ModuleNotFoundError: No module named 'sklearn'`

See https://pypi.org/project/sklearn/


JackCaoG (approved):
Thanks!

PR #743, opened 2022-11-14
Clean up unnecessary TF tests.

will-cromar:
Remove >130 tests.

Includes many legacy models that were already replaced at the time of release (e. g. `resnet-ctl` and `classifier-resnet` vs `vision-imagenet`).  Whatever error in our process included those in the 2.9 and 2.10 releases seems to have been solved in 2.11 -- those tests match nightly.

Disabling 2.10 tests since we're close to a 2.11 release (currently on rc2).

FYI @chandrasekhard2


aman2930:
Thanks Will for cleaning these up.

PR #745, opened 2022-11-14
Updating the schedule for TF-2.11 tests and removing OLD TF test targets.

aman2930:


tests/tensorflow/targets.jsonnet
```diff
@@ -22,9 +22,5 @@ local r2_9 = import 'r2.9/targets.jsonnet';
 // Add new versions here
 std.flattenArrays([
   nightly,
-  r2_7,
```
will-cromar, line 25: Please revert this file -- #743 already removes these tests.

aman2930: I think you just changed the schedule. While I was trying not to push these workloads to GKE. Let me know if that makes sense.

PR #747, opened 2022-11-15
Added NLP BERT DATA DIR

chandrasekhard2:


RissyRan:
Could you check this file (https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/k8s/europe-west4/gcs-buckets.yaml) and add it in the same order? Thanks!

RissyRan (approved):
LGTM!

PR #755, opened 2022-11-30
[PT/XLA] Add DDP test configs for ResNet50

will-cromar:
Use `pytorch/xla/test/pjrt/test_train_pjrt_imagenet.py` and write summaries for all tests (PJRT and XRT) now that blockers have been removed. Needs follow-up PR to add parameter broadcasting that test to fix correctness on TPU v3 (WIP).


will-cromar:
Tested some of the new configs with manual runs.

will-cromar:
CC @alanwaketan

PR #772, opened 2022-12-21
Adding profiler test for WMT model, adding dependencies for WMT

RoshaniN:
1. Adding a profiler test - "flax-latest-wmt-profile-v3-8-1vm" which runs the WMT model and verifies the existence of a trace.
2. Including WMT model dependencies, i.e. correct version of protobuf ( to avoid incompatibility issues with TF ) and tensorflow_datasets.

( Issues are not related to this XL-ML test directly, but dependencies of WMT model itself.

Protobuf 3.20.* was needed to avoid tensorflow import error-
"TypeError: Descriptors cannot not be created directly.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower)."
 
 tensorflow_datasets import was needed because tensorflow_datasets is also imported in WMT. )

### DIFF
--- a/tests/jax/latest/wmt.libsonnet
+++ b/tests/jax/latest/wmt.libsonnet
@@ -25,6 +25,11 @@ local tpus = import 'templates/tpus.libsonnet';
     extraConfig:: 'default.py',
     extraFlags+:: ['--config.reverse_translation=True', '--config.per_device_batch_size=32'],
   },
+  local profile = mixins.Functional {
+    mode: 'profile',
+    extraFlags+:: ['--config.num_train_steps=40', '--config.per_device_batch_size=16'],
+    extraConfig:: 'default.py',
+  },
   local v3_8 = {
     accelerator: tpus.v3_8,
   },
@@ -33,10 +38,17 @@ local tpus = import 'templates/tpus.libsonnet';
   },
   local wmt = common.runFlaxLatest {
     modelName:: 'wmt',
-    extraDeps+:: ['tensorflow_text sentencepiece'],
+    extraDeps+:: ['tensorflow_text sentencepiece', 'protobuf==3.20.*', 'tensorflow_datasets'],
+  },
+  local wmt_profiling = wmt {
+    local config = self,
+    testScript+:: |||
+      gsutil -q stat $(MODEL_DIR)/plugins/profile/*/*.xplane.pb
+    ||| % (self.scriptConfig {}),
   },
   configs: [
     wmt + functional + v2_8,
     wmt + convergence + v3_8 + timeouts.Hours(20),
+    wmt_profiling + profile + v3_8 + timeouts.Hours(1),
   ],
 }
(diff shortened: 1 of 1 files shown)


RoshaniN:
https://buganizer.corp.google.com/issues/263413201 - logged a bug for the dependency issues.

RoshaniN:
@will-cromar Could you take a look ?

will-cromar (approved):

tests/jax/latest/wmt.libsonnet
```diff
@@ -33,10 +38,17 @@ local tpus = import 'templates/tpus.libsonnet';
...
+  local wmt_profiling = wmt {
+    local config = self,
+    testScript+:: |||
+      gsutil -q stat $(MODEL_DIR)/plugins/profile/*/*.xplane.pb
+    ||| % (self.scriptConfig {}),
   },
   configs: [
     wmt + functional + v2_8,
     wmt + convergence + v3_8 + timeouts.Hours(20),
+    wmt_profiling + profile + v3_8 + timeouts.Hours(1),
```
bvandermoon, line 52: How long has the profiler test been taking to run? Just wondering if the timeout could be cut to something like 20-30 minutes instead

tests/jax/latest/wmt.libsonnet
```diff
@@ -33,10 +38,17 @@ local tpus = import 'templates/tpus.libsonnet';
   },
   local wmt = common.runFlaxLatest {
     modelName:: 'wmt',
-    extraDeps+:: ['tensorflow_text sentencepiece'],
+    extraDeps+:: ['tensorflow_text sentencepiece', 'protobuf==3.20.*', 'tensorflow_datasets'],
```
bvandermoon, line 41: Minor: seems like two formatting styles are being mixed here. tensorflow_text and sentencepiece are separated by space and the others are separated by commas

bvandermoon (approved):

tests/jax/latest/wmt.libsonnet
RoshaniN: Takes about 15-20 mins as per my observation. I didn't want to be too strict with the timeout here, as sometimes it takes ~10 mins to just provision the resources. 
There is an option to reduce it to 20-30 minutes - https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/templates/timeouts.libsonnet#L23 . We could explore those in the next set of changes too.

tests/jax/latest/wmt.libsonnet
RoshaniN: I agree, I plan to remove the dependencies that has been added here soon after the bug is fixed.

bvandermoon (approved):

PR #781, opened 2023-01-10
Refactor async broadcast

gobbleturk:
Use multihost_utils.broadcast_one_to_all

Tested via run_multipod_test.sh
(Tested 5x for consistency)


rwitten:
Looks great! Can you test it more like 100+ times? (I imagine you can either test it multiple times per invocation of run_multipod_test or put a loop around run_multipod_test.)

gobbleturk:
Ran it 100 times, all passed =)

PR #782, opened 2023-01-11
Remove directory creation

gobbleturk:
Remove directory creation from the synchronous orbax test, since orbax has been updated to this automatically.
Also remove the option to clean the directory after the test since we are setting it to false anyway

This passed 100 runs on `run_multipod_test.sh`

Previously this could flake if the second host tried to create the `checkpointManager` before the first host created the directory. This could have been fixed with a `multihost_utils.sync_global_devices()` after the directory creation, but now the explicit directory creation is not needed and in fact orbax does a `multihost_utils.sync_global_devices()` internally after directory creation.


rwitten (approved):
LGTM, thanks!

PR #787, opened 2023-01-19
Update python-ops.libsonnet to install expecttest

wonjoolee95:
Since we moved python-op tests to 1VM, tests are failing with:
```
Traceback (most recent call last):
  File "/home/xl-ml-test/pytorch/xla/test/../../test/test_view_ops.py", line 11, in <module>
    from torch.testing._internal.common_utils import (
  File "/home/xl-ml-test/.local/lib/python3.8/site-packages/torch/testing/_internal/common_utils.py", line 57, in <module>
    import expecttest
ModuleNotFoundError: No module named 'expecttest'
```
This PR adds install statement to `expecttest`.

### DIFF
--- a/tests/pytorch/nightly/python-ops.libsonnet
+++ b/tests/pytorch/nightly/python-ops.libsonnet
@@ -47,6 +47,7 @@ local utils = import 'templates/utils.libsonnet';
       tpuVmExtraSetup: |||
         echo 'export PATH=~/.local/bin:$PATH' >> ~/.bash_profile
         echo 'export XLA_USE_BF16=1' >> ~/.bash_profile
+        pip install expecttest
       |||,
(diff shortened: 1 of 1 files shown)


will-cromar (approved):
This is fine as a temporary fix.

Should we add an [`extras_require`](https://github.com/pytorch/xla/blob/5c54655a8d43f48d973b664c20afaa91c7278574/setup.py#L388) in PT/XLA for common test dependencies like this, `tensorboardX`, etc?

wonjoolee95:
Sounds good, I'll merge this for now. We do have an open PR from Manfei -- https://github.com/pytorch/xla/pull/4444 that might address this. If it does, I'll revert this PR.

PR #792, opened 2023-01-25
added importlib-metadata as a requirement for metrics handler

gkroiz:
Deploying metrics handler with bazelisk ran into an error without importlib-metadata.


will-cromar (approved):
Thanks!

PR #793, opened 2023-01-26
[PT/XLA] Reorganize DDP tests

will-cromar:
- Set `torch.distributed` env vars for XRT DDP tests
- Remove tests for custom PJRT DDP
- Add a convergence test with DDP
- Re-enable some PJRT tests and disable some XRT tests

Depends on https://github.com/pytorch/xla/pull/4520

### DIFF
--- a/tests/pytorch/nightly/mnist.libsonnet
+++ b/tests/pytorch/nightly/mnist.libsonnet
@@ -96,11 +96,19 @@ local utils = import 'templates/utils.libsonnet';
       modelDir: null,
     },
   },
+  local pjrt_ddp = {
+    modelName+: '-ddp',
+    command+: [
+      '--ddp',
+      '--pjrt_distributed',
+    ],
+  },
 
   configs: [
     mnist + convergence + v2_8 + timeouts.Hours(1) + mixins.Experimental,
     mnist + convergence + v2_8 + timeouts.Hours(1) + tpuVm + mixins.Experimental,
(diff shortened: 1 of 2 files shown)


tests/pytorch/nightly/mnist.libsonnet
```diff
@@ -96,11 +96,19 @@ local utils = import 'templates/utils.libsonnet';
...
+      '--ddp',
+      '--pjrt_distributed',
+    ],
+  },
 
   configs: [
     mnist + convergence + v2_8 + timeouts.Hours(1) + mixins.Experimental,
     mnist + convergence + v2_8 + timeouts.Hours(1) + tpuVm + mixins.Experimental,
-    mnist + convergence + v2_8 + timeouts.Hours(1) + pjrt + mixins.Experimental,
+    mnist + convergence + v2_8 + timeouts.Hours(1) + pjrt,
```
JackCaoG, line 110: do we want to add mnist conv test back? Would resnet be enough?

JackCaoG (approved):

tests/pytorch/nightly/mnist.libsonnet
will-cromar: I added this one back since it's our only e2e test that works on TPU v2 that I know of. I can take it off of the schedule if there's another test that covers PJRT + TPU v2.

JackCaoG: we can keep it.

PR #798, opened 2023-02-08
[PT/XLA] Move end-to-end tests to PJRT C API

will-cromar:
Fix FS transformer test by switching back to original XRT code and re-enabled PJRT tests.

Started all 3 PJRT models (ResNet50, FS Transformer, and DLRM) manually to confirm they start training without crashing.


JackCaoG:
Please keep an eye on tmr's test.

PR #802, opened 2023-02-13
Adjusted BigQuery schema and restructured related queries

gkroiz:


gkroiz:
@jiangjy1982 @will-cromar

dashboard/main_heatmap.py
```diff
@@ -34,51 +34,73 @@
...
-  test_name LIKE @test_name_prefix AND
-  (metric_lower_bound IS NOT NULL OR metric_upper_bound IS NOT NULL) AND
-  (metric_value < metric_lower_bound OR metric_value > metric_upper_bound) AND
+  job.test_name,
+  SAFE_CAST(DATE(job.timestamp, 'US/Pacific') AS STRING) AS run_date,
+  metrics.metric_name,
+  metrics.metric_value,
+  metrics.metric_upper_bound,
+  metrics.metric_lower_bound
+FROM (SELECT *
```
jiangjy1982, line 88: Nit: Keep the style consistent:
```
FROM (
  SELECT
```
Also try not to use the wildcard *, but spell the fields out explicitly.

dashboard/main_heatmap.py
```diff
@@ -34,51 +34,73 @@
...
+  FROM `{METRIC_HISTORY_TABLE_NAME}`
+  WHERE (metric_lower_bound IS NOT NULL OR metric_upper_bound IS NOT NULL) AND
+  (metric_value <= metric_lower_bound OR metric_value >= metric_upper_bound)
+  ) AS metrics
+INNER JOIN
```
jiangjy1982, line 93: Will there be cases where a uuid appears in metrics but not job_history or vice versa? If it's possible, we should consider surfacing such issues by using outer joins.

dashboard/metric_compare.py
```diff
@@ -29,47 +29,54 @@
...
-    x.timestamp,
-    x.metric_value,
-    x.uuid
-  FROM (
-    SELECT
+  SELECT 
+    test_name,
+    job_status,
+    timestamp,
+    FIRST_VALUE (timestamp) OVER (
```
jiangjy1982, line 44: Again, can be further simplified by using `row_number()`.

jiangjy1982 (changes requested):

dashboard/main_heatmap.py
```diff
@@ -34,51 +34,73 @@
...
-  y.test_name = x.test_name AND
-  y.max_uuid = farm_fingerprint(x.uuid)
-ORDER BY
-  run_date DESC
+  )
+  ORDER BY
+    timestamp DESC
+)
+WHERE timestamp = first_time
 """
```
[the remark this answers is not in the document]
will-cromar: We can get this even more simple. Here's what I have:

```
SELECT 
  test_name,
  job_status,
  SAFE_CAST(DATE(timestamp, 'US/Pacific') AS STRING) as run_date,
  stackdriver_logs_link,
  kubernetes_workload_link,
  uuid,
  RANK() OVER (
    PARTITION BY
        test_name, DATE(timestamp, 'US/Pacific')
    ORDER BY 
        timestamp DESC
  ) as rnk
FROM `xl-ml-test.metrics_handler_dataset.job_history`
WHERE DATE(timestamp) >= DATE('2023-01-01') AND test_name like 'pt-nightly-%'
QUALIFY rnk = 1
```

I tried moving the `WHERE` clauses into a subquery to filter before partitioning, but that didn't impact query performance, so I think BQ might apply that optimization automatically while letting us keep the query clean.

dashboard/main_heatmap.py
will-cromar: I actually think it's okay to use `*` in a subquery like this, since we clearly enumerate the columns in the final output. That is, unless using `*` affects performance, but I don't think it would in this specific instance.

dashboard/main_heatmap.py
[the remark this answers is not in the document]
jiangjy1982: Ha, I didn't know BQ also has the QUALIFY keyword. Yes, your query beats mine :)

dashboard/main_heatmap.py
```diff
@@ -34,51 +34,73 @@
...
-  SELECT
+SELECT 
+  test_name,
+  job_status,
+  SAFE_CAST(DATE(timestamp, 'US/Pacific') AS STRING) as run_date,
+  logs_link,
+  workload_link,
+  uuid
+FROM
+(SELECT
```
RissyRan, line 45: Nit: you may want to keep the query format consistent in those files. i.e. location of `(`.

RissyRan (changes requested):
I think Will's proposal is simpler. I assume you will/have tested those changes in the test database right?

gkroiz:
Will close PR and recreate two separate PRs, one for producer side and one for consumer side. Will add changes once tf and pytorch releases are finished.

PR #807, opened 2023-02-20
Added configuration so that 2VM TF 2.12 tests use TF version 2.12

gkroiz:
Since the 2vm tests cannot find tf software version 2.12.0, we have to use tf 2.11.0, and then switch to tf 2.12.0 manually. This has to be done for both red and green vm

### DIFF
--- a/tests/tensorflow/r2.12/common.libsonnet
+++ b/tests/tensorflow/r2.12/common.libsonnet
@@ -23,9 +23,70 @@ local mixins = import 'templates/mixins.libsonnet';
 
     frameworkPrefix: 'tf-r2.12.0',
     tpuSettings+: {
-      softwareVersion: '2.12.0',
+      softwareVersion: '2.11.0',
     },
     imageTag: 'r2.12.0',
+    podTemplate+:: {
+      spec+: {
+        initContainerMap+:: {
+          'tpu-version': {
+            image: config.podTemplate.spec.containerMap.train.image,
+            env+: [
+              {
+                name: 'TPU_NAME',
+                valueFrom: {
+                  fieldRef: {
+                    fieldPath: "metadata.annotations['name.cloud-tpus.google.com/train']",
+                  },
+                },
+              },
+              {
+                name: 'POD_UID',
+                valueFrom: {
+                  fieldRef: {
+                    fieldPath: 'metadata.uid',
+                  },
+                },
+              },
+            ],
+            local tpuCreateSettings = {
+              acceleratorName: std.escapeStringBash(config.accelerator.name),
+              softwareVersion: std.escapeStringBash(config.tpuSettings.softwareVersion),
+              startupScript: std.escapeStringBash(config.tpuSettings.tpuVmStartupScript),
+              sleepTime: config.tpuSettings.tpuVmCreateSleepSeconds,
+              testName: std.strReplace(config.testName, '.', '-'),
+            },
+            command: [
+              'python3',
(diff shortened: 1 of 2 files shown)


tests/tensorflow/r2.12/common.libsonnet
```diff
@@ -46,7 +99,7 @@ local mixins = import 'templates/mixins.libsonnet';
       runnerPath: 'official/nlp/train.py',
     },
   },
-  local functional_schedule = '0 3 * * *',
+  local functional_schedule = '0 7 * * *',
```
chandrasekhard2, line 102: I changed the tests to avoid conflict with nightly tests. Please revert this.

tests/tensorflow/r2.12/common.libsonnet
```diff
@@ -74,7 +127,7 @@ local mixins = import 'templates/mixins.libsonnet';
   },
   // Override default schedule for Functional.
   RunNightly:: {
-    schedule: '0 5 * * *',
+    schedule: functional_schedule,
```
chandrasekhard2, line 130: ditto

gkroiz: I reverted and repushed without these changes

tests/tensorflow/r2.12/common.libsonnet
gkroiz: I reverted and repushed without these changes

tests/tensorflow/r2.12/common.libsonnet
```diff
@@ -23,9 +23,62 @@ local mixins = import 'templates/mixins.libsonnet';
 
     frameworkPrefix: 'tf-r2.12.0',
     tpuSettings+: {
-      softwareVersion: '2.12.0',
+      softwareVersion: '2.11.0',
```
chandrasekhard2, line 26: This can now be 2.12.0. Can you manually trigger a build and check if its working properly?

gkroiz: Setting this to 2.12.0 will still result in an error because the debian image is not in prod yet.

tests/tensorflow/r2.12/common.libsonnet
```diff
@@ -23,9 +23,62 @@ local mixins = import 'templates/mixins.libsonnet';
...
+              },
+            ],
+            local tpuCreateSettings = {
+              acceleratorName: std.escapeStringBash(config.accelerator.name),
+              softwareVersion: std.escapeStringBash(config.tpuSettings.softwareVersion),
+              startupScript: std.escapeStringBash(config.tpuSettings.tpuVmStartupScript),
+              sleepTime: config.tpuSettings.tpuVmCreateSleepSeconds,
+              testName: std.strReplace(config.testName, '.', '-'),
+            },
+            command: [
```
allenwang28, line 59: Would this prepend an existing command, or overwrite it altogether? Or does that even matter in this context?

gkroiz: I belive this would overwrite an existing command but to my knowledge there is not one that exists, so it does not matter. The change in this PR is similar to what pytorch uses for 2vm: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/pytorch/nightly/common.libsonnet#L47-L73

aman2930 (approved):

chandrasekhard2 (approved):

PR #813, opened 2023-02-22
add --no-index to libtpu install command

aireenmei:
see context in b/270151193. Tested on my TPU VM


will-cromar (approved):

will-cromar:
This LGTM in principle, but the diffs are empty. Can you confirm this code path is being used?

will-cromar:
I think the diff check is broken. Merging.

PR #817, opened 2023-02-23
Remove k8s/**/gen from .gitignore

will-cromar:
The is preventing any changes from being reflected on the `gen` branch.


gkroiz (approved):

gkroiz:
Just for reference. The line in .gitignore is being removed because it is preventing the generated files in k8s from being commited to gen, which is needed for the tests to run.

PR #836, opened 2023-03-14
Add script to schedule tests

steventk-g:
# Description

Adds a script to schedule a batch of oneshots for one type of test, based on matching the start of the jobname. For example, if we want to run all of the `pt-nightly` tests, we can run `XLML_TEST_TYPE=pt-nightly ./scripts/schedule_tests.sh`. It connects to every cluster on which we have tests, gathers the tests from those clusters, and creates the jobs.

# Tests

This can be used for any test, but doesn't modify the test itself.

Run something like `XLML_TEST_TYPE=pt-nightly ./scripts/schedule_tests.sh`

Example: http://shortn/_FEDtYR413y

# Checklist

### DIFF
--- a/scripts/schedule_tests.sh
+++ b/scripts/schedule_tests.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+if [[ -z "${XLML_TEST_TYPE}" ]]; then
+    echo "Please set XLML_TEST_TYPE"
+    exit 1
+else
+    test_type="${XLML_TEST_TYPE}"
+fi
+
+regions=(us-central1 us-central2 europe-west4)
+for region in ${regions[@]}; do
+
+    gcloud container clusters get-credentials xl-ml-test-"$region" --region "$region" --project xl-ml-test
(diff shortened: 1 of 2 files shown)


scripts/schedule_tests.sh
```diff
@@ -0,0 +1,30 @@
...
+fi
+
+# Loop over regions and connect to each cluster
+regions=(us-central1 us-central2 europe-west4)
+for region in ${regions[@]}; do
+    gcloud container clusters get-credentials xl-ml-test-"$region" --region "$region" --project xl-ml-test
+
+    # Loop over cron jobs, grepping for the test type
+    for i in $(kubectl get cronjobs -n automated -o name | grep "$test_type"); do
+        echo "$i"
```
jiangjy1982, line 21: How do we prevent accidentally kick off many jobs matching to a short/trivial $test_type? Shall we add a confirmation step?

steventk-g: Good call, I'll add a confirmation prompt before scheduling any jobs.

will-cromar (approved):
This is great! Thanks!

PR #849, opened 2023-03-27
Refactor tests to support adding this repository as a submodule

will-cromar:
# Description

Make it easier to import this repository as a submodule and add new test variants. Expose most test configuration as fields (rather than locals) to make them accessible from other files. I left the existing `local`s in place to make the `configs` section less verbose (e.g. `mnist + funtional` vs `self.mnist + self.functional`).

# Tests

Test run with PT/XLA MNIST example: http://shortn/_lFUC1nIJZJ
Test run with TF example (currently failing on dashboard too): http://shortn/_I1A8T3wAbW

Diffs have no changes except for whitespace.

# Checklist

### DIFF
--- a/tests/pytorch/nightly/hf-glue.libsonnet
+++ b/tests/pytorch/nightly/hf-glue.libsonnet
@@ -29,7 +29,8 @@ local utils = import 'templates/utils.libsonnet';
   local command_copy_metrics = |||
     gsutil -m cp -r ./tensorboard-metrics/* $(MODEL_DIR)
   |||,
-  local bert_base_cased = common.Convergence {
+  local bert_base_cased = self.bert_base_cased,
+  bert_base_cased:: common.Convergence {
     modelName: 'hf-glue-bert-b-c',
     paramsOverride+:: {
       model_name_or_path: 'bert-base-cased',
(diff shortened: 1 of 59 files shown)


tests/pytorch/nightly/hf-glue.libsonnet
```diff
@@ -87,9 +89,9 @@ local utils = import 'templates/utils.libsonnet';
     },
     command: utils.scriptCommand(
       |||
-        %s 
-        %s 
-        %s 
+        %s
+        %s
+        %s
```
jonb377, line 94: The diff looks fine to me, but I didn't read through line-by-line. Since this should be causing most of the diff, should we move this out of the current PR to verify the resulting diff is empty?

jonb377 (approved):
LGTM, thanks Will! One small comment about the diff, but don't treat it as blocking.

tests/pytorch/nightly/hf-glue.libsonnet
will-cromar: My editor must have removed these by accident. I checked each of the diffs, and the only change is this whitespace, plus the change I made to use gcloud ssh in the TF tests.

PR #845, opened 2023-03-20
bq query optimizations (consumer side)

gkroiz:
# Description
Consumer side BQ queries optimizations for xl ml dashboard. These changes were made to improve query time while maintaining/improving readability. More details [here](https://tinyurl.com/2979uh88). 

# Tests
- Tests within `dashboard/` passed.
- To check that the logic of the new queries matches the old queries, I tested the queries against each other using different test names, metric names and over different periods of time. From each of these comparisons, the number of entries matched, the queried rows matched, and the ordering matched when ordering mattered.
- Ran dashboard locally and tested functionality for heatmap, metrics plotting, and metrics compare

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [X] I have performed a self-review of my code.
- [X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X] I have run one-shot tests and provided workload links above if applicable. 
- [X] I have made or will make corresponding changes to the doc if needed.

PR #851, opened 2023-03-28
removed logs_download_command references

gkroiz:
# Description
There is a bug with the current dashboard where clicking on a test does not show additional test information. Culprit change is #845. This PR fixes this bug.

# Tests
Hosted dashboard locally via
```
python3 -m bokeh serve --show dashboard/dashboard.py dashboard/metrics.py
```
and compared dashboard and metrics with live dashboard. Also now able to click on individual tests

# Checklist

### DIFF
--- a/dashboard/javascript_utils.py
+++ b/dashboard/javascript_utils.py
@@ -98,14 +98,6 @@ PASS_FAIL_GRID_MODAL_STRING = """
       workload_link_html_element = `<br/><br/><a href="${workload_link}" target="_blank">Kubernetes Workload</a>`;
     }
 
-    var logs_download_html = '';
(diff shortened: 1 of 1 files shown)


jiangjy1982:
So the fix is to remove codes rather than bringing back codes around `logs_download_command`?

gkroiz:
> So the fix is to remove codes rather than bringing back codes around `logs_download_command`?

Yes, PR #845 removed `logs_download_command` as a returned attribute from the queries. `logs_download_command` is an outdated attribute that is no longer populated, so we no longer need to query for it. Rather than adding back in this attribute in the queries for the fix, I removed where `logs_download_command` is referenced within the dashboard UI.

will-cromar (approved):

PR #859, opened 2023-04-05
Fixes bug in tensorflow Dockerfile

EricLeFort:
# Description

This fixes a bug in the TensorFlow Dockerfile. The file is currently throwing exceptions because it does not have valid `/bin/sh` syntax and because variables were out of scope in places they were being used. This change fixes both issues.

# Tests

This doesn't affect tests, just the image generation script used to run them. I ran the image generation script locally using:
```
sudo docker build . \
-f images/tensorflow/Dockerfile \
--build-arg image_version=${_BASE_IMAGE_VERSION} \
--build-arg model_garden_branch=${_MODEL_GARDEN_BRANCH} \
--build-arg tensorflow_text_version=${_TENSORFLOW_TEXT_VERSION}
```

# Checklist

### DIFF
--- a/images/tensorflow/Dockerfile
+++ b/images/tensorflow/Dockerfile
@@ -12,23 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG image_version=nightly
+ARG image_version="nightly"
+
 FROM tensorflow/tensorflow:${image_version}-gpu
 
-ARG model_garden_branch=master
+ARG image_version
+ARG model_garden_branch="master"
+ARG tensorflow_text_version="nightly"
 
 RUN apt-get update && apt-get install -y --no-install-recommends git
-
 RUN pip3 install cloud-tpu-client pyyaml
-
-ARG tensorflow_text_version=nightly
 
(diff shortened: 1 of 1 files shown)


will-cromar (approved):
Thanks!

images/tensorflow/Dockerfile
```diff
@@ -12,23 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG image_version=nightly
+ARG image_version="nightly"
+
 FROM tensorflow/tensorflow:${image_version}-gpu
```
chandrasekhard2, line 17: Current RedVM has python:3.9.16-bullseye. Therefore this need to be

FROM python:3.9.16-bullseye

images/tensorflow/Dockerfile
```diff
@@ -12,23 +12,25 @@
...
```
chandrasekhard2, line 18: Manually install tensorflow-gpu pip package in python:3.9.16-bullseye image.

`RUN pip install tensorflow-gpu`

PR #860, opened 2023-04-07
Refactor TF nightly tests

chandrasekhard2:
# Description

Refactored tf-nightly tests

# Tests

TF-nightly tests

./scripts/run-oneshot.sh -t tf.nightly-retinanet-coco-func-v4-8-1vm

retinanet test - http://shortn/_0ngoVolUmC

# Checklist


RissyRan:
Thanks for the change. Is the test supposed to pass?

chandrasekhard2:
> 

Hi Ran, The test is for checking if the right image is being pulled. I verified that by logging into the TPU manually while the test is running. But the tests fail due to API changes in the nightly build. Another fix will be coming soon.This is more of a fix for stale tests.

RissyRan:
Thanks, Chandra! LGTM!

Nit: you could `put X in square brackets` instead of start (*), so those items will be checked.

RissyRan (approved):

tests/tensorflow/nightly/base.libsonnet
```diff
@@ -0,0 +1,195 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+local utils = import 'templates/utils.libsonnet';
+local volumes = import 'templates/volumes.libsonnet';
+
+{
+  BaseTpuVmTest:: {
```
will-cromar, line 19: This file essentially copies [experimental.libsonnet](https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/experimental.libsonnet). These classes should instead override `experimental.BaseTpuVmTest` where necessary.

tests/tensorflow/nightly/tf-retinanet-coco.libsonnet
```diff
@@ -39,11 +39,13 @@ local utils = import 'templates/utils.libsonnet';
     },
   },
   local functional = common.Functional {
-    paramsOverride+: {
-      trainer+: {
-        train_steps: 400,
-        validation_interval: 200,
-        validation_steps: 100,
+    scriptConfig+: {
```
will-cromar, line 42: Good catch

tests/tensorflow/nightly/experimental.libsonnet
```diff
@@ -0,0 +1,127 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+local experimental = import 'base.libsonnet';
+local mixins = import 'templates/mixins.libsonnet';
+
+{
+  TensorFlowTpuVmMixin:: experimental.BaseTpuVmMixin + mixins.Unsuspended {
```
will-cromar, line 19: Why is this file copied here from `tensorflow/nightly`?

https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/tests/tensorflow/experimental.libsonnet

tests/tensorflow/nightly/common.libsonnet
```diff
@@ -25,6 +26,87 @@ local mixins = import 'templates/mixins.libsonnet';
...
+                '-c',
+                |||
+                  import os
+                  import tensorflow as tf
+                  import urllib
+                  import json
+                  import cloud_tpu_client
+                  import sys
+                  print('python version: ' + str(sys.version))
+                  print('tf_version: ' + str(tf.__version__))
```
will-cromar, line 71: This could be a good place to fail out early if the image is too old. Of course, this would only work for TPU Node.

cc @jiangjy1982

will-cromar (changes requested):

tests/tensorflow/nightly/common.libsonnet
jiangjy1982: Thanks @will-cromar for the thorough review! @chandrasekhard2 Let's put a line `TODO(chandrasekhard2): Fail early if the image is stale` here and defer to a future PR so that we don't dilute the purpose of this one.

tests/tensorflow/nightly/base.libsonnet
chandrasekhard2: Yeah I tried to copy and separate out tensorflow tests. But I think thats not a good idea. I removed the copy and change it to experimental.BaseTpuVmTest. Thanks!

tests/tensorflow/nightly/experimental.libsonnet
chandrasekhard2: I wanted to separate out all the relevant files into tensorflow

tests/tensorflow/nightly/common.libsonnet
chandrasekhard2: Done. Thanks!

chandrasekhard2:
> Thanks, Chandra! LGTM!
> 
> Nit: you could `put X in square brackets` instead of start (*), so those items will be checked.

Done. Thanks!

dashboard/app.yaml
```diff
@@ -33,7 +33,7 @@ resources:
 env_variables:
   REDISHOST: '10.25.27.107'
   REDISPORT: '6379'
-  TEST_NAME_PREFIXES: 'pt-nightly,pt-2.0,tf.nightly,tf.r2.12.0,tf.r2.11.1,%-1vm,jax,flax,pax,mp-jax,mp-pax'
```
will-cromar, line 36: Thanks for cleaning up the old tests. Can you separate this into a new PR to keep this one focused on the nightly tests?

tests/experimental.libsonnet
```diff
@@ -96,33 +93,49 @@ local volumes = import 'templates/volumes.libsonnet';
...
-                set -e
-                test $exit_code = 0 && break || sleep 30;
+              curl -X POST \
+                -H "Authorization: Bearer $(gcloud auth print-access-token)" \
+                -H "Content-Type: application/json" \
+                -d "{
+                  accelerator_type: %(acceleratorName)s,
+                  runtime_version: %(softwareVersion)s,
+                  network_config: {enable_external_ips: true},
+                  boot_disk: {source_image: 'projects/cloud-tpu-v2-images-dev/global/images/family/tpu-vm-tf-nightly'},
```
will-cromar, line 103: Please move this `curl` logic into `tensorflow/`'s `common.TpuVm`. This change will affect all frameworks. We should prefer to use `gcloud` where we can, and only use `curl` where we need to set the boot disk override.

tests/tensorflow/nightly/common.libsonnet
```diff
@@ -129,11 +213,7 @@ local mixins = import 'templates/mixins.libsonnet';
   },
   local functional_schedule = '0 9 * * 3',
   Functional:: mixins.Functional {
-    schedule:
-      if !(self.accelerator.type == 'tpu') || self.accelerator.name == 'v3-8' || self.accelerator.name == 'v4-8' then
-        functional_schedule
-      else
-        null,
+    schedule: functional_schedule,
```
will-cromar, line 216: This will re-enable a number of currently disabled tests that currently have their schedules set to `null` here, e.g. `tf.nightly-resnet-imagenet-func-v2-8.yaml`

I'm fine removing this logic (it's really not very intuitive), but please either remove the currently-disabled test configs or mark them with `+ mixins.Experimental` (this will set the schedule to `null`).

tests/tensorflow/nightly/tf-dcn-criteo.libsonnet
```diff
@@ -94,6 +94,18 @@ local tpus = import 'templates/tpus.libsonnet';
     },
   },
 
+  local v2_32 = tpu_common {
```
will-cromar, line 97: What is this change for?

will-cromar (changes requested):
Please review the `diff` GitHub Actions output. This PR currently has a much larger "blast radius" than you intended (e.g. changing TPU creating for every TPU VM test).

tests/experimental.libsonnet
chandrasekhard2: Yeah I added that in the latest commit.

tests/tensorflow/nightly/common.libsonnet
chandrasekhard2: I think its better to leave this as is.

tests/tensorflow/nightly/tf-dcn-criteo.libsonnet
chandrasekhard2: I wanted to keep this for config reference for eg., what bottom_mlp and embedding_dim would work on v2_32.

tests/tensorflow/nightly/tf-resnet-imagenet.libsonnet
```diff
@@ -15,6 +15,7 @@
 local experimental = import '../experimental.libsonnet';
 local common = import 'common.libsonnet';
 local mixins = import 'templates/mixins.libsonnet';
+local timeouts = import 'templates/timeouts.libsonnet';
```
will-cromar, line 18: Is this addition accidental?

tests/tensorflow/targets.jsonnet
```diff
@@ -19,6 +19,5 @@ local r2_12 = import 'r2.12/targets.jsonnet';
 // Add new versions here
 std.flattenArrays([
   nightly,
-  r2_11,
```
will-cromar, line 22: Please move this to your other PR as well.

will-cromar (commented):
Thanks. These changes look good in principle, but I would like to check the diffs. Can you run `scripts/format-jsonnet.sh`?

tests/tensorflow/nightly/tf-resnet-imagenet.libsonnet
chandrasekhard2: Oh thanks Will for catching it. It is accidental and used no where in the file.

chandrasekhard2:
> Thanks. These changes look good in principle, but I would like to check the diffs. Can you run `scripts/format-jsonnet.sh`?

It passes all checks now. Thanks @will-cromar . I will send another PR for cleaning old tests.

tests/tensorflow/nightly/common.libsonnet
```diff
@@ -25,6 +28,174 @@ local mixins = import 'templates/mixins.libsonnet';
...
+                |||,
+              ],
+            },
+          },
+        },
+      }
+    else
+      {},
+  },
+  tpuVm:: experimental.TensorFlowTpuVmMixin {
```
will-cromar, line 96: Sorry I missed this on my last pass: you should be able to eliminate all fo the overrides here except for `podTemplate.spec.initContainerMap['create-tpu'].command`, since that's the only thing that's changing from the parent.

chandrasekhard2: Actually I tried that but the local variables like tpuCreateSettings is not reflected here.  There are other unbound variable errors that I encountered.

will-cromar: You won't be able to take out the `local` variables if they're using in `command` like `tpuCreateSettings`, since they're scoped to this object. You can take out e.g. `podTemplate.spec.initContainerMap['create-tpu'].image` since that will be inherited

chandrasekhard2: Removing that line results in an error - 

The Job "tf.nightly-maskrcnn-coco-func-v4-8-1vm-jbrz7" is invalid: spec.template.spec.initContainers[0].image: Required value

will-cromar: I see use `'create-tpu'+:` (note the `+`) to override individual fields rather than replace all of them.

tests/tensorflow/nightly/common.libsonnet
```diff
@@ -25,6 +28,174 @@ local mixins = import 'templates/mixins.libsonnet';
...
+              " > /scripts/cleanup.sh
+
+              curl -X POST \
+                -H "Authorization: Bearer $(gcloud auth print-access-token)" \
+                -H "Content-Type: application/json" \
+                -d "{
+                  accelerator_type: %(acceleratorName)s,
+                  runtime_version: %(softwareVersion)s,
+                  network_config: {enable_external_ips: true},
+                  boot_disk: {source_image: 'projects/cloud-tpu-v2-images-dev/global/images/family/tpu-vm-tf-nightly'},
```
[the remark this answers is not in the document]
chandrasekhard2: Added

tests/tensorflow/nightly/common.libsonnet
chandrasekhard2: Done

PR #862, opened 2023-04-10
Update TPU name format

RissyRan:
# Description

Add `versionName` as a string since `version` field is used to calculate core number.

# Tests

**Instruction and/or command lines to reproduce your tests:** 

```
./scripts/run-oneshot.sh -t flax.latest-bart-wiki.summary-func-v4-8
```

**List links for your tests (use go/shortn-gen for any internal link):** 

The test is successful: [link](http://shortn/_8E7QsHQw64)

# Checklist

### DIFF
--- a/templates/tpus.libsonnet
+++ b/templates/tpus.libsonnet
@@ -18,10 +18,10 @@ local base = import 'base.libsonnet';
   TpuSpec:: base.BaseAccelerator {
     local tpu = self,
 
-    name: 'v%(versionName)s-%(size)d' % tpu,
+    name: 'v%(version)d%(variant)s-%(size)d' % tpu,
     type: 'tpu',
     version: error 'Must override `version`',
-    versionName: error 'Must override `versionName`',
+    variant: error 'Must override `variant`',
     size: error 'Must override `size`',
     numCores: if tpu.version <= 3 then 8 else 4,
     replicas: tpu.size / 8,  // Each TPU replica has 8 cores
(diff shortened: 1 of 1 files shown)


templates/tpus.libsonnet
```diff
@@ -18,9 +18,10 @@ local base = import 'base.libsonnet';
   TpuSpec:: base.BaseAccelerator {
     local tpu = self,
 
-    name: 'v%(version)d-%(size)d' % tpu,
+    name: 'v%(versionName)s-%(size)d' % tpu,
     type: 'tpu',
     version: error 'Must override `version`',
+    versionName: error 'Must override `versionName`',
```
will-cromar, line 24: Can this default to `str(self.version)` for the common case?

jonb377: Alternatively, could we just change the type of `version` to a string instead of adding a new field? It looks like it's only used in this file.

RissyRan: Probably not, due to this line `numCores: if tpu.version <= 3 then 8 else 4`.

RissyRan: Hmm, it seems the number is being used in pytorch. https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fml-testing-accelerators%20numCores&type=code

will-cromar: Being able to check version ranges is helpful for e.g. finding how many devices there are.

What about adding another field `variant`? Then `name` would be like `v%(version)d%(variant)s-%(size)d`, e.g. `v4-8` or `v4abcd-8`

RissyRan: I like this one, and will update it.

RissyRan: Tested it via http://shortn/_kvfc52UbDD. Also, tested internally.

will-cromar (approved):

PR #864, opened 2023-04-12
Remove tf 2.11 tests and disable tf-2.12 tests

chandrasekhard2:
# Description

Remove TF 2.11 tests as they're no longer required and disable TF 2.12 tests.

# Tests

N/A

N/A

N/A

# Checklist


tests/tensorflow/targets.jsonnet
```diff
@@ -13,12 +13,9 @@
...
 
 local nightly = import 'nightly/targets.jsonnet';
-local r2_11 = import 'r2.11/targets.jsonnet';
 local r2_12 = import 'r2.12/targets.jsonnet';
 
 // Add new versions here
 std.flattenArrays([
   nightly,
-  r2_11,
-  r2_12,
```
will-cromar, line 23: nit: if you remove `r2_12` here, then you should remove the whole `r2.12` folder since they won't be in the build. Otherwise, setting `schedule: null` and making sure there are no overrides should be sufficient

chandrasekhard2: Done. Thanks!

will-cromar (approved):
Thanks!

PR #891, opened 2023-05-08
Remove default duration assertion

will-cromar:
# Description

After #884, `duration` is no longer a meaningful metric. The `Job`'s active time may be much longer than the test's actual duration.

# Tests

Will double-check diffs.

# Checklist

### DIFF
--- a/tests/common.libsonnet
+++ b/tests/common.libsonnet
@@ -42,18 +42,6 @@ local metrics = import 'templates/metrics.libsonnet';
           ],
           merge_runs: false,
         },
-        literals: {
-          assertions: {
-            duration: {
-              inclusive_bounds: false,
-              std_devs_from_mean: {
-                comparison: 'LESS',
-                std_devs: 5,
-              },
-              wait_for_n_data_points: 10,
-            },
-          },
-        },
       },
     },
 


RissyRan:
Would like to confirm after the change, there is no timeout (duration assertion) before the pod is active? And after the pod is active, the timeout is max of [24h, 2 x specified timeout].

will-cromar:
This only removes the metric assertion on the job duration. The actual timeout will only happen if (1) the pod is actually created and takes longer than the specified timeout or (2) the Job has waited a very long time to create the pod and still failed.

RissyRan:
`(1) the pod is actually created and takes longer than the specified timeout `: is this the max of [24h, 2 x specified timeout] from this [change](https://github.com/GoogleCloudPlatform/ml-testing-accelerators/pull/884)?

`(2) the Job has waited a very long time to create the pod and still failed.`: what is the timeout for this one? 10 min?

will-cromar:
(1) is the value of the `timeout` field in the test (e.g. set by `timeouts.Hours(2)`) and (2) is the [24h, 2 x specified timeout] number from #884.

RissyRan (approved):
Thanks!

PR #893, opened 2023-05-09
Add script to clean up unused TPU VMs

will-cromar:
# Description

Hacky script that deletes any TPU that does not correspond to a running Pod in a known cluster/namespace.

Move the run-oneshots.sh script off of the oneshots-* clusters. It makes this sort of management easier, and the deployment triggers already update the `default` namespace in those clusters.

Once submitted, I can add a trigger to run this every ~15 minutes to clean up any unowned TPUs automatically.

# Tests

With updated oneshot script: http://shortn/_T3g1OeS8t7

Ran `tpu-cleanup.sh` manually. It found a leaked TPU and ignored currently used and non-test TPU VMs.

# Checklist

### DIFF
--- a/scripts/tpu-cleanup.sh
+++ b/scripts/tpu-cleanup.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+set -x
+set -u
+
+zones=("us-central1-b" "us-central2-b" "europe-west4-a")
+
+for zone in ${zones[@]}; do
+  region=${zone%-*}
+  cluster=xl-ml-test-$region
+  gcloud container clusters get-credentials $cluster --region $region --project xl-ml-test
+
+  # Don't delete any TPUs for pods that are currently running
+  pods_to_save=$(mktemp)
+  for namespace in default automated; do
+    for status in Running Pending; do
+      kubectl get pods -n $namespace --field-selector status.phase=$status -o jsonpath='{.items[*].metadata.uid}' \
+        | tr ' ' '\n' \
+          >> $pods_to_save
+      # Ensure there is a newline between outputs
+      echo "" >> $pods_to_save
+    done
+  done
+
+  # Strip empty lines that could confuse grep
+  sed -i -ne/./p $pods_to_save
+
+  leaks=$(
+    gcloud compute tpus tpu-vm list --zone $zone --project=xl-ml-test --format="value(name)" \
+      | grep "^tpu-" \
+      | grep -v -f $pods_to_save)
+  for leak in $leaks; do
+    gcloud compute tpus tpu-vm delete --quiet --async --zone $zone --project=xl-ml-test $leak
+  done
+done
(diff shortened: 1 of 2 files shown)


scripts/tpu-cleanup.sh
```diff
@@ -0,0 +1,34 @@
...
+
+zones=("us-central1-b" "us-central2-b" "europe-west4-a")
+
+for zone in ${zones[@]}; do
+  region=${zone%-*}
+  cluster=xl-ml-test-$region
+  gcloud container clusters get-credentials $cluster --region $region --project xl-ml-test
+
+  # Don't delete any TPUs for pods that are currently running
+  pods_to_save=$(mktemp)
```
vanbasten23, line 13: I wonder what `$(mktemp)` is

scripts/tpu-cleanup.sh
```diff
@@ -0,0 +1,34 @@
...
+  for namespace in default automated; do
+    for status in Running Pending; do
```
RissyRan, line 15: How do we distinguish if the status of TPUs is actually running? I see you comment `Don't delete any TPUs for pods that are currently running` but here includes `Running`. What's the difference?

scripts/tpu-cleanup.sh
RissyRan: I think it makes a temp file and write to-be-deleted TPUs there.

scripts/tpu-cleanup.sh
will-cromar: Basically, we make a list of every Kubernetes pod that might still be using a TPU, and then exclude any TPU associated with those pods. We can map from TPU VM to Kubernetes pod easily, because the TPUs are named like `tpu-${POD_UID}`

scripts/tpu-cleanup.sh
```diff
@@ -0,0 +1,34 @@
...
+      kubectl get pods -n $namespace --field-selector status.phase=$status -o jsonpath='{.items[*].metadata.uid}' \
+        | tr ' ' '\n' \
+          >> $pods_to_save
+      # Ensure there is a newline between outputs
+      echo "" >> $pods_to_save
```
jonb377, line 20: At the cost of readability,  `sed -ie '$a\' $pods_to_save` will append a newline if absent. Then line 25 could be removed IIUC

https://stackoverflow.com/questions/23055831/add-new-line-character-at-the-end-of-file-in-bash

jonb377 (approved):
Is the link in the description to the demo run correct?

Otherwise LGTM, thanks Will!

scripts/tpu-cleanup.sh
will-cromar: Clarified this comment.

will-cromar:
Good find! Done.

will-cromar:
The link in the description is a test run with the updated oneshot script. Clarified the description.

scripts/tpu-cleanup.sh
RissyRan: Thanks! LGTM!

RissyRan (approved):

PR #896, opened 2023-05-11
removing unused import that breaks the test

sshahrokhi:
# Description

An unused import was causing the test to fail.

# Tests

I ran it locally to make sure the cache tests are written. 

And, here is the link to the one shot run I ran: http://shortn/_gKmaXWHuIV

# Checklist


RissyRan (approved):
Thanks!

PR #899, opened 2023-05-15
Removed tpu node tests for tf nightly tests

gkroiz:
# Description
This PR removes all existing tf-nightly tests that use TPU nodes.
# Tests
Compare `./scripts/list-tests.sh` before and after removing the tests:
http://shortn/_kua7eNcIAL
The link above shows that the there are no remaining tpu node tests and all of the tpu vm and gpu tests are still included.

# Checklist


tests/tensorflow/nightly/targets.jsonnet
```diff
@@ -12,23 +12,19 @@
...
 local resnetrs = import 'tf-resnetrs-imagenet.libsonnet';
 local retinanet = import 'tf-retinanet-coco.libsonnet';
-local wmt = import 'tf-wmt-wmt14_translate.libsonnet';
 
 // Add new models here
 std.flattenArrays([
   dlrm.configs,
   dcn.configs,
   keras_api.configs,
-  bert.configs,
```
RissyRan, line 30: Instead of deleting those tests in bert and wmt models, shall we just convert 2vm to 1vm? So that we are still monitoring these 2 models. Based on HuggingFace downloads:
* wmt is top model in Translation task 
* bert is top model in Fill-Mask task

gkroiz: Would TPU VM capacity be a concern? If these models are still relevant I think it is a good idea to convert some of the TPU Node tests to TPU VM tests. Should we only convert the tests for wmt and bert, or is it worthwhile converting all of the TPU Node tests to TPU VM tests.

will-cromar: We should have coverage over some NLP task. We can/should make room in the capacity by dropping some of the similar models. E.g. retinanet or mask-rcnn, resnet or resnetrs

gkroiz: Yes, but AFAIK, there is not much consistency between which generation of TPUs we test for which tests, and in some cases we will only have a convergence test for X-generation of TPU but not a functional test. Unless there is a structure to this, in the future we should try to organize which generations are being used for which tests.

gkroiz: Added back wmt and bert as 1vm tests

RissyRan: Thanks! Could you trigger a functional test to ensure 2 models are working expected?

gkroiz: Assuming you are referring to bert and wmt, I'll run a functional test for each of these and will update once they finish

gkroiz: Update here, I was blocked until this morning, both the bert and wmt tests are failing due to a SIGABRT. IMO we should merge this PR regardless (to free up 2vm resources) and track these failing test separately.

RissyRan: Discussed offline. I agree to free up 2vm resources first, and create a bug to track the fix. Approving.

RissyRan (approved):

PR #923, opened 2023-06-21
Extending .gitignore to ignore generated tests

EricLeFort:
# Description

Adds a .gitignore line for the generated tests since they are quite distracting when working through diffs.

Since the workflow still needs those files, I also added a line to force add them there.

# Tests

N/A, no test changes. I will monitor the dashboard after merging to ensure nothing has been broken.

**Instruction and/or command lines to reproduce your tests:** N/A

**List links for your tests (use go/shortn-gen for any internal link):** N/A

# Checklist

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -33,8 +33,9 @@ jobs:
         scripts/gen-tests.sh
     
     - name: Git Commit/Push Changes
-      uses: actions-x/commit@v5
-      with:
-        branch: gen
-        files: k8s/
-        force: true
+      run: |
+        git add --force -N k8s/*/gen/*.yaml
+        git add -N k8s/
+        git commit -m "Automatically updated using GitHub Actions"
+        git push -f origin gen
+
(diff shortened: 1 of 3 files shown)


will-cromar:
Can you update this action too? https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/master/.github/workflows/generate.yml

EricLeFort:
Wasn't sure if there was a way to get that pre-built action to do what we needed so made it a custom run similar to `diff.yml`.

.github/workflows/generate.yml
```diff
@@ -33,8 +33,9 @@ jobs:
...
-      uses: actions-x/commit@v5
-      with:
-        branch: gen
-        files: k8s/
-        force: true
+      run: |
+        git add --force -N k8s/*/gen/*.yaml
+        git add -N k8s/
+        git commit -m "Automatically updated using GitHub Actions"
+        git push -f
```
will-cromar, line 40: Should this be `push -f origin gen`? I don't think the upstream is set here.

EricLeFort: That's safer regardless, will change.

will-cromar (approved):
Thanks!

PR #925, opened 2023-06-22
Bugfix: removing -N flag from git commands in generate.yaml

EricLeFort:
# Description

The `-N` flag adds the files to the diff but doesn't actually add them to be committed, this is not what we want here.

# Tests

N/A

**Instruction and/or command lines to reproduce your tests:** N/A

**List links for your tests (use go/shortn-gen for any internal link):** N/A

# Checklist

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -37,8 +37,8 @@ jobs:
         # Add the generated tests using the most recent authors name/email
         git config --global user.name "`git log -1 --pretty=format:'%an'`"
         git config --global user.email "`git log -1 --pretty=format:'%ae'`"
-        git add --force -N k8s/*/gen/*.yaml
-        git add -N k8s/
+        git add --force k8s/*/gen/*.yaml
+        git add k8s/
         git commit -m "Automatically updated using GitHub Actions"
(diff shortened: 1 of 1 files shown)


.github/workflows/generate.yml
```diff
@@ -37,8 +37,8 @@ jobs:
         # Add the generated tests using the most recent authors name/email
         git config --global user.name "`git log -1 --pretty=format:'%an'`"
         git config --global user.email "`git log -1 --pretty=format:'%ae'`"
-        git add --force -N k8s/*/gen/*.yaml
-        git add -N k8s/
+        git add --force k8s/*/gen/*.yaml
+        git add k8s/
```
will-cromar, line 41: Do we need both of these adds? The only new files in the `gen` branch should be at the paths above.

will-cromar (approved):

.github/workflows/generate.yml
EricLeFort: In practice, no.
In theory, this makes it a little more robust in case some new file gets added there that should also be included.

It does no harm though (I double-checked locally that it doesn't un-add those files or anything).

PR #946, opened 2023-07-09
Revert "update PTL sd test" and update installation

gkroiz:
Reverts #944 and updates PTD SD dependencies directly.

#944 changed how packages were installed, using `requirements.txt` in https://github.com/pytorch-tpu/stable-diffusion rather than direct pip installations. The main issue with this is that the requirements in https://github.com/pytorch-tpu/stable-diffusion depend on the latest stable release for Pytorch and Pytorch XLA. To keep up with the nightly versions of Pytorch and Pytorch XLA, we will need to make changes that override `requirements.txt`. As such, I think it's best to revert this previous PR and instead update dependencies directly.

### DIFF
--- a/tests/pytorch/nightly/sd-model.libsonnet
+++ b/tests/pytorch/nightly/sd-model.libsonnet
@@ -51,15 +51,22 @@ local utils = import 'templates/utils.libsonnet';
         pip install transformers==4.19.2 diffusers invisible-watermark
         pip install -e .
         pip install torchmetrics==0.7.0
-        pip install https://github.com/PyTorchLightning/pytorch-lightning/archive/master.zip
+        pip install https://github.com/Lightning-AI/lightning/archive/refs/heads/master.zip -U
         pip install lmdb einops omegaconf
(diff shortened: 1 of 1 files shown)


gkroiz:
> I noticed there are some updates compared to the original [file](https://github.com/GoogleCloudPlatform/ml-testing-accelerators/pull/944/files), is this expected during the "revert"?
> 
> * original `starlette==0.22.0` while in the revert version is `starlette==0.27.0 `?
> * original `pip install https://github.com/PyTorchLightning/pytorch-lightning/archive/master.zip`, while in the revert version is `pip install https://github.com/Lightning-AI/lightning/archive/refs/heads/master.zip -U`?
> 
> If you have made changes accordingly, let's call this PR something like `Revert "update PTL sd test" and update installation`?

Updated the PR title.

gkroiz:
Here is the oneshot: http://shortn/_UuK3K8PT0p

RissyRan (approved):

PR #954, opened 2023-07-13
Upgrading Python version to 3.10 in tensorflow Dockerfile

EricLeFort:
# Description

Modified the tensorflow red VM image Dockerfile to upgrade the Python version from 3.8 to 3.10

This resolves a bug caused by [this line](https://github.com/tensorflow/models/blob/master/official/core/base_trainer.py#L457) which uses dict union syntax only available in Python 3.10+.

The plan will be to simply revert these changes once the TensorFlow team upgrades the base image to use 22.04 (which comes with Python 3.10).

# Tests

The following tests verify the fix. The convergence test is failing during eval using head, the functional test was already successful on head and still is.

**Instruction and/or command lines to reproduce your tests:**
```
./scripts/run-oneshot.sh -t tf.nightly-dlrm-criteo-func-v100-x1
./scripts/run-oneshot.sh -t tf.nightly-dlrm-criteo-conv-v100-x1
```

**List links for your tests (use go/shortn-gen for any internal link):**
Func results: http://shortn/_1slyPfh2iU
Conv results: http://shortn/_ONBuEpBvZT

# Checklist

### DIFF
--- a/images/tensorflow/Dockerfile
+++ b/images/tensorflow/Dockerfile
@@ -20,7 +20,16 @@ ARG image_version
 ARG model_garden_branch="master"
 ARG tensorflow_text_version="nightly"
 
-RUN apt-get update && apt-get install -y --no-install-recommends git
+# TODO (ericlefort) remove this once the base image gets updated to 22.04
+# Keep track of the full set of dependencies in 3.8 and reinstall them on 3.10
+RUN pip3 freeze > requirements.txt
+RUN apt-get install software-properties-common && apt-get update && add-apt-repository ppa:deadsnakes/ppa
+RUN apt-get update && apt-get install -y --no-install-recommends git python3.10 python3.10-distutils python3.10-dev libcairo2-dev
+RUN cd /usr/bin &&  unlink python3 && ln -s /usr/bin/python3.10 python3
+RUN curl -sS https://bootstrap.pypa.io/get-pip.py | python3.10
+RUN cat requirements.txt | xargs -n 1 pip3 install
+RUN rm requirements.txt
+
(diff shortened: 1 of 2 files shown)


images/tensorflow/Dockerfile
```diff
@@ -20,7 +20,16 @@ ARG image_version
 ARG model_garden_branch="master"
 ARG tensorflow_text_version="nightly"
 
-RUN apt-get update && apt-get install -y --no-install-recommends git
+# TODO (ericlefort) remove this once the base image gets updated to 22.04
+# Keep track of the full set of dependencies in 3.8 and reinstall them on 3.10
+RUN pip3 freeze > requirements.txt
+RUN apt-get install software-properties-common && apt-get update && add-apt-repository ppa:deadsnakes/ppa
```
will-cromar, line 26: If you're going to reinstall everything, then I strongly suggest starting from the `python:3.10` base image, or an `nvidia/cuda` base image. IMO, the real benefit of using the upstream TF image off of the shelf if that it is tested before we get it on Dockerhub. If you're creating a new python env within that image, I'm not sure how helpful it really is.

will-cromar (commented):
~What's the motivation for this change?~

Edit: never mind, I should have read the description

will-cromar (approved):
LGTM as a short term fix if the TF model garden isn't willing to support Python 3.8

images/tensorflow/Dockerfile
EricLeFort: Yeah, it's definitely not pretty. We do expect the TF team to do this upgrade in the near-term though (still confirming a plan/timeline with them) so it's only a band-aid until then.

PR #1036, opened 2024-01-03
update wheel build from linux to manylinux

zpcore:
Since the `*-linux_*` wheel didn't get updated when we re-trigger the build, we switch to the `manylinux` version, which gets updated every time we re-trigger the build.


will-cromar (approved):
Good catch!

I wonder if we should switch to an actual index here so we don't have to update the exact file name like this.

ManfeiBai (approved):
LGTM

ManfeiBai:
> Good catch!
> 
> I wonder if we should switch to an actual index here so we don't have to update the exact file name like this.

what does the `switch to an actual index` here mean? feel that's could be a good solution too

will merge this PR now to enable rc6 test and let's continue to the solution in the follow-up PR

PR #1045, opened 2024-02-09
Update PJRT_DEVICE to TPU

zpcore:
xl-ml pytorch nightly tests failed due to we are still using `TPU_C_API`.


ManfeiBai (commented):
Thanks! and let's share the one-shot run link here too

will-cromar (approved):

